PR #2, opened 2021-04-28
Recommendations without fallbackFilters shouldn't return random results

Shipow:
As of today, the `hitsPerPage` prop is driving the number of recommendation we display; whatever the number of recommendation we actually get and whether or not you have fallback filters.

Especially for Bought Together; you rather don't want any results than random hits, should you not have any recommendations.

### DIFF
--- a/src/Recommendations.js
+++ b/src/Recommendations.js
@@ -39,7 +39,6 @@ const buildSearchParamsFromRecommendations_TEMPORARY_BETA = (record, props) => {
   let recoFilters = [];
   let hitsPerPage = props.hitsPerPage;
   const threshold = props.threshold || 0;
-  const model = props.model;
 
   if (record.recommendations) {
     recoFilters = record.recommendations
@@ -54,7 +53,7 @@ const buildSearchParamsFromRecommendations_TEMPORARY_BETA = (record, props) => {
       hitsPerPage = record.recommendations.length;
     }
   }
-  if (model === 'bought-together') {
+  if (this.model === 'bought-together') {
     return {
       optionalFilters: [],
       filters:
--- a/src/Recommendations.js
+++ b/src/Recommendations.js
@@ -129,10 +129,10 @@ export class Recommendations extends Component {
   }
 
   render() {
-    if (this.state.recommendations.length === 0) {
+    if (this.props.model ==="bought-together" && this.state.recommendations.length === 0) {
       return (
         <div className="text-indigo-500 text-sm text-italic">
-          No match in recommend model {this.props.model}
+          Nothing yet...
         </div>
       );
     }
--- a/src/Recommendations.js
+++ b/src/Recommendations.js
@@ -33,10 +33,19 @@ const getRecommendedObject_TEMPORARY_BETA = (
     });
 };
 
+const composeFilters = (recoFilters) => {
+  let compose = '';
+  if (recoFilters.length > 0) {
+    compose += `(${recoFilters.join(' OR ')})`;
+  }
+  return compose;
+};
+
 const buildSearchParamsFromRecommendations_TEMPORARY_BETA = (record, props) => {
   let recoFilters = [];
   let hitsPerPage = props.hitsPerPage;
   const threshold = props.threshold || 0;
+  const model = props.model;
 
   if (record.recommendations) {
     recoFilters = record.recommendations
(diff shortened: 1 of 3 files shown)


src/Recommendations.js
```diff
@@ -33,10 +33,19 @@ const getRecommendedObject_TEMPORARY_BETA = (
     });
 };
 
+const composeFilters = (recoFilters) => {
+  let compose = '';
+  if (recoFilters.length > 0) {
+    compose += `(${recoFilters.join(' OR ')})`;
+  }
+  return compose;
+};
```
dispix, line 42: This is equivalent to:

```suggestion
const composeFilters = (recoFilters) => recoFilters.join(' OR ');
```

Shipow: yes thanks, I didn't clean after removing the negative filter from the recipe!

src/Recommendations.js
```diff
@@ -51,13 +54,24 @@ const buildSearchParamsFromRecommendations_TEMPORARY_BETA = (record, props) => {
...
+      hitsPerPage,
+    };
+  } else if (model === 'related-products') {
+    return {
+      optionalFilters: [...recoFilters, ...(props.fallbackFilters || [])],
+      filters: `NOT objectID:${props.objectID}`,
+      facetFilters: props.facetFilters,
+      hitsPerPage,
+    };
+  }
```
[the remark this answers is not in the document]
Kosai106: I just read the PR name - Why would we not want to fallback onto a specific model?

src/Recommendations.js
```diff
@@ -109,6 +121,13 @@ export class Recommendations extends Component {
   }
 
   render() {
+    if (this.props.model ==="bought-together" && this.state.recommendations.length === 0) {
+      return (
+        <div className="text-indigo-500 text-sm text-italic">
+          Nothing yet...
```
redox, line 127: We cannot hardcode those classes & (english) text in the component.

src/Recommendations.js
```diff
@@ -51,12 +53,22 @@ const buildSearchParamsFromRecommendations_TEMPORARY_BETA = (record, props) => {
       hitsPerPage = record.recommendations.length;
     }
   }
-
+  if (this.model === 'bought-together') {
```
redox, line 56:
```suggestion
  if (props.model === 'bought-together') {
```

src/Recommendations.js
```diff
@@ -51,12 +53,22 @@ const buildSearchParamsFromRecommendations_TEMPORARY_BETA = (record, props) => {
...
+    return {
+      optionalFilters: [],
+      filters:
+        (recoFilters.length > 0 &&
+          composeFilters(recoFilters)) ||
+        null,
```
redox, line 62:
```suggestion
      filters: composeFilters(recoFilters) || null,
```

The `recoFilters.length > 0` test should already be covered by your logic above; isn't it?

src/Recommendations.js
```diff
@@ -51,12 +53,22 @@ const buildSearchParamsFromRecommendations_TEMPORARY_BETA = (record, props) => {
...
+      facetFilters: (recoFilters.length > 0 && props.facetFilters) || [],
```
redox, line 63: I don't get why you would only do the filtering if there are some recommentations.

redox (changes requested):

src/Recommendations.js
```diff
@@ -33,30 +33,49 @@ const getRecommendedObject_TEMPORARY_BETA = (
...
+    recoFilters = [];
+
+    // no recommendations but fallback, force to retrieve maxRecommendations hits
+    if (hasFallback) {
+      hitsPerPage = maxRecommendations;
+    } else {
+      // otherwise, don't retrieve anything
+      hitsPerPage = 0;
     }
   }
```
redox, line 72: This is where the main change is. What do you think @Shipow @mthuret ?

src/Recommendations.js
```diff
@@ -33,30 +33,56 @@ const getRecommendedObject_TEMPORARY_BETA = (
...
```
redox, line 79: This is where the main change is. What do you think @Shipow @mthuret ?

mthuret: Works for me :)

Shipow: lgtm for now!
I'll explore a wrapper props options on another branch.

PR #4, opened 2021-04-30
feat(widget): wrapperComponent

Shipow:
add a props to use an advanced wrapper (ie: a carousel)


Shipow:
100x better implem on next by @francoischalifour

PR #7, opened 2021-05-14
feat(userAgent): Add `recommendations` user agent

shortcuts:
**Summary**

Adds user agent to the `useRecommendations` hook.

**Result**
When submitting a new query, the following user agents will be added:
- React: `react-recommendations (version)`
- JavaScript: `react-recommendations (version)`

### DIFF
--- a/packages/react-recommendations/src/useRecommendations.ts
+++ b/packages/react-recommendations/src/useRecommendations.ts
@@ -12,6 +12,7 @@ import {
   getIndexNameFromModel,
   getOptionalFilters,
 } from './utils';
+import { version } from './version';
 
 export type UseRecommendationsProps = {
   model: RecommendationModel;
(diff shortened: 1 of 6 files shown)


packages/react-recommendations/src/useRecommendations.ts
```diff
@@ -63,6 +64,14 @@ export function useRecommendations<TObject extends ProductBaseRecord>(
       .then((record) => {
         const recommendations = record.recommendations ?? [];
 
+        if (
+          !props.searchClient.transporter.userAgent.value.includes(
+            `js-recommendations (${version})`
+          )
+        ) {
+          props.searchClient.addAlgoliaAgent('react-recommendations', version);
+        }
```
francoischalifour: We don't need to add the UA at every query, but we can rather add it every time the search client changes–in an effect with `props.searchClient` as dependencies.

shortcuts: Done in https://github.com/algolia/react-recommendations/pull/7/commits/5d2720b33ba359eadf4c477dde72d4b2617302d9

francoischalifour (approved):

PR #8, opened 2021-05-14
feat: accept multiple `objectID`s

francoischalifour:
## Description

In the context of a shopping cart, getting recommendations based on all items is a common pattern. This updates the API to accept multiple `objectID`s as references to get recommendations from.

## How it works

- We get recommendations for each object ID passed to `objectIDs` via `searchClient.getObjects`, with a `hitsPerPage` value divided by the number of `objectIDs`
- We then build an Algolia query for each of these recommendations
- The recommendations can contain duplicate `objectID`s, so we remove any duplications.
- We sort the recommendations by their scores to merge the recommendations coming from all the `objectIDs` in a continuous way

## Questions

- We might want to sort before deduping so that the lowest scored version of the item gets removed?

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,7 +30,7 @@ Then, use the `Recommendations` component:
   model="related-products"
   searchClient={searchClient}
   indexName="YOUR_SOURCE_INDEX_NAME"
-  objectID={objectID}
+  objectIDs={[objectID]}
   hitComponent={Hit}
 />
 
@@ -39,7 +39,7 @@ Then, use the `Recommendations` component:
   model="bought-together"
   searchClient={searchClient}
   indexName="YOUR_SOURCE_INDEX_NAME"
-  objectID={objectID}
+  objectIDs={[objectID]}
   hitComponent={Hit}
 />
 ```
@@ -64,11 +64,11 @@ The initialized Algolia search client.
 
 The name of the products index.
 
-### `objectID`
+### `objectIDs`
 
-> `string` | **required**
+> `string[` | **required**
 
-The objectID of the product to get recommendations from
+An array of object IDs of the products to get recommendations from.
 
 ### `hitComponent`
 
--- a/README.md
+++ b/README.md
@@ -66,9 +66,9 @@ The name of the products index.
 
 ### `objectIDs`
 
-> `string[` | **required**
+> `string[]` | **required**
 
-An array of object IDs of the products to get recommendations from.
+An array of `objectID`s of the products to get recommendations from.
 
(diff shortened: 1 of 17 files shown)


README.md
```diff
@@ -64,11 +64,11 @@ The initialized Algolia search client.
 
 The name of the products index.
 
-### `objectID`
+### `objectIDs`
 
-> `string` | **required**
+> `string[` | **required**
```
shortcuts:
```suggestion
> `string[]` | **required**
```

README.md
```diff
@@ -64,11 +64,11 @@ The initialized Algolia search client.
...
 
-The objectID of the product to get recommendations from
+An array of object IDs of the products to get recommendations from.
```
shortcuts:
```suggestion
An array of objectIDs of the products to get recommendations from.
```

Shipow (approved):
thanks!!!

packages/react-recommendations/src/utils/sortBy.ts
```diff
@@ -0,0 +1,8 @@
+type Predicate<TITem> = (a: TITem, b: TITem) => number;
```
shortcuts: We should probably stick to `TObject` since it's already implemented in the codebase, wdyt?

(Otherwise, there's a typo in the type `TITem` -> `TItem`)

shortcuts (approved):
> We might want to sort before deduping so that the lowest scored version of the item gets removed?

If it's possible to have 2 recommend score for the same objectID, yes

packages/react-recommendations/src/utils/sortBy.ts
francoischalifour: These utils are supposed to be used with any data structure so there's no strict reason to use the same naming as the generic Algolia records. But it should be cased `TItem` indeed.

shortcuts: Make sense!

francoischalifour: See f54ef351295225772dfb819f15bfdc1691af6b14.

README.md
francoischalifour: See 984ced12f9dbc55b47f9bd4150ea5a625d9f8047.

README.md
francoischalifour: See 984ced12f9dbc55b47f9bd4150ea5a625d9f8047.

francoischalifour:
@shortcuts I noted it on my board—I'll take care of deduping correctly in another PR along with some tests.

PR #9, opened 2021-05-14
feat(transformItems): Add `transformItems` function to `useRecommendations` hook

shortcuts:
**Summary**

Similar API as in [InstantSearch.js](https://www.algolia.com/doc/api-reference/widgets/hits/js/#widget-param-transformitems), it receive the `items` returned by the query and allow the user to change its content, order, etc. The shape returned should remain the same.

**Results**

Any `recommendations` component can now use `transformItems` to transform the results returned by Algolia. 

**Preview**
<p align="center">
  
</p>


packages/react-recommendations/src/useRecommendations.ts
```diff
@@ -23,15 +24,16 @@ export type UseRecommendationsProps = {
   maxRecommendations?: number;
   searchParameters?: SearchOptions;
   threshold?: number;
+  transformItems?: RecommendationTransformItems<TObject>;
```
francoischalifour: We can inline `RecommendationTransformItems` if we only use it here.

francoischalifour (approved):
Let's merge this once https://github.com/algolia/react-recommendations/pull/9/files#r632578219 is addressed, if it makes sense to you.

packages/react-recommendations/src/types/RecommendationTransformItems.ts
```diff
@@ -0,0 +1,3 @@
+export type RecommendationTransformItems<TObject> = (
```
Haroenv: why is this not called `TransformItems`, as it's not related to recommendations

francoischalifour: We can inline it so we don't have to name it.

packages/react-recommendations/src/useRecommendations.ts
shortcuts: Sure!

packages/react-recommendations/src/types/RecommendationTransformItems.ts
shortcuts: Done in https://github.com/algolia/react-recommendations/pull/9/commits/c4e34b585b80325f2f55bef15c2d68451683c7cb

packages/react-recommendations/src/useRecommendations.ts
shortcuts: Done in https://github.com/algolia/react-recommendations/pull/9/commits/c4e34b585b80325f2f55bef15c2d68451683c7cb

Haroenv (approved):

francoischalifour (approved):

PR #10, opened 2021-05-14
docs(examples): support demo config

francoischalifour:
This adds a demo configuration to change credentials and attributes based on `.env`. We need this to iterate on private indices.

We still default the demo config for CodeSandbox usage.

From now on, you can add a `.env` file at the monorepo root and change the demo credentials.


francoischalifour:
Closing this PR because we don't need this anymore.

PR #13, opened 2021-05-19
feat: introduce slider components

francoischalifour:
## Description

This extracts the slider component from the AI components to plug it with a `view` prop. The default view is a basic list view.

Slider Preview

This `view` prop is not yet compatible with the JavaScript package, I'll address that in another PR.

### Related products for React

```tsx
import { RelatedProducts } from '@algolia/react-recommendations';
import { HorizontalSlider } from '@algolia/ui-components-horizontal-slider-react';
import '@algolia/ui-components-horizontal-slider-react/HorizontalSlider.css';

function App() {
  return (
    <RelatedProducts
      searchClient={searchClient}
      indexName={indexName}
      objectIDs={[selectedProduct.objectID]}
      hitComponent={Hit}
      view={HorizontalSlider}
    />
  );
}
```

You can use the same view with `FrequentlyBoughtTogether`.

### Standalone `HorizontalSlider` component for React

```tsx
import { HorizontalSlider } from '@algolia/ui-components-horizontal-slider-react';
import '@algolia/ui-components-horizontal-slider-react/HorizontalSlider.css';

function App() {
  return <HorizontalSlider itemComponent={itemComponent} items={items} />;
}
```

### Standalone `horizontalSlider` component for JavaScript

```tsx
import { horizontalSlider } from '@algolia/ui-components-horizontal-slider-react';
import '@algolia/ui-components-horizontal-slider-react/HorizontalSlider.css';

horizontalSlider({ itemComponent, items });
```

## New packages

- `@algolia/ui-components-horizontal-slider-js`
- `@algolia/ui-components-horizontal-slider-react`

## Changes

This PR also required some changes in the Babel and Rollup configuration, because React was not correctly aliased to Preact in the JS package.

## Next steps

- Document the new packages
- Support `view` param in the JavaScript Recommendations package
- Extract the Horizontal Slider theme to a package
- Find better naming for component and packages?


packages/horizontal-slider-react/src/HorizontalSlider.tsx
```diff
@@ -0,0 +1,183 @@
...
+  root: string;
+};
+
+type HorizontalSliderTranslations = {
+  sliderLabel: string;
+  previousButtonLabel: string;
+  previousButtonTitle: string;
+  nextButtonLabel: string;
+  nextButtonTitle: string;
+};
```
shortcuts: Why don't we export those types?

If we do so, it should be encapsulated with `Partial<{...}>`, otherwise, the user won't have the "real" type and will have to do it on his side

packages/horizontal-slider-react/src/HorizontalSlider.tsx
```diff
@@ -0,0 +1,183 @@
...
+            scrollRight();
+          }
+        }}
+      >
+        {props.items.map((item, index) => (
+          <li
+            key={item.objectID}
+            className={cx('uic-HorizontalSlider-item', props.classNames?.item)}
+            aria-roledescription="slide"
+            aria-label={`${index + 1} of ${props.items.length}`}
```
shortcuts, line 150: We might want to provide this word as a translations, wdyt? 

Maybe something like `sliderItemLabel` and provide a small caption with an example, defaulted to `of`

shortcuts (commented):
Looks really good and promising! 

- this file should be deleted: `examples/vanilla-demo/app.js`
- I think we need to update [these](https://github.com/algolia/ui-components/blob/next/.circleci/config.yml#L57-L64) and [these](https://github.com/algolia/ui-components/blob/next/.circleci/config.yml#L23-L30) lines with the two new packages to make the CI work again

packages/horizontal-slider-react/src/HorizontalSlider.tsx
```diff
@@ -0,0 +1,183 @@
...
+  objectID: string;
+};
+
+type HorizontalSliderClassnames = {
+  item: string;
+  list: string;
+  navigation: string;
+  navigationNext: string;
+  navigationPrevious: string;
+  root: string;
```
shortcuts, line 23:
```suggestion
  item?: string;
  list?: string;
  navigation?: string;
  navigationNext?: string;
  navigationPrevious?: string;
  root?: string;
```

with `const { classNames = {} } = props;` L#44 would avoid having the `props.classNames?.` syntax

packages/horizontal-slider-react/src/HorizontalSlider.tsx
```diff
@@ -0,0 +1,183 @@
...
+    >
+      <button
+        ref={previousButtonRef}
+        title={translations.previousButtonTitle}
+        aria-label={translations.previousButtonLabel}
+        aria-controls={sliderIdRef.current}
+        className={cx(
+          'uic-HorizontalSlider-navigation',
+          'uic-HorizontalSlider-navigation--previous',
+          props.classNames?.navigation,
```
Haroenv: maybe classnames should be defaulted to {} to avoid checking at every usage?

packages/horizontal-slider-react/src/HorizontalSlider.tsx
```diff
@@ -0,0 +1,183 @@
...
+        }}
+      >
+        {props.items.map((item, index) => (
+          <li
+            key={item.objectID}
+            className={cx('uic-HorizontalSlider-item', props.classNames?.item)}
+            aria-roledescription="slide"
+            aria-label={`${index + 1} of ${props.items.length}`}
+          >
+            <props.itemComponent item={item} />
```
Haroenv, line 152: isn't a component meant to start with a capital letter? If it doesn't error it's not really a problem I guess

Haroenv (commented):
hard to tell whether this is sufficient, but it at least looks reasonable!

packages/horizontal-slider-react/src/HorizontalSlider.tsx
Haroenv: if we have this as a translation, it likely would need to be completely a translation, since the order won't be the same in every language, and we didn't accept functions as translations. Not sure what's the solution

shortcuts: That's true but it would cover most cases until we encounter this issue, I guess it's better than having English by default

shortcuts: An other/complementary solution for this could be to provide a kind-of RTF boolean variable that would swap the places of the two numbers if setted

francoischalifour: That's tricky because that should be a function indeed.

We can still introduce this is a function translation later because we need to release today. Wdyt?

packages/horizontal-slider-react/src/HorizontalSlider.tsx
francoischalifour: We went for a similar pattern as `hitComponent` in React InstantSearch, but I'm not sold TBH.

Do you have alternatives in mind?

packages/horizontal-slider-react/src/HorizontalSlider.tsx
francoischalifour: See 17462ff5628938a3592e4d77da0bab7fff5b6497.

packages/horizontal-slider-react/src/HorizontalSlider.tsx
francoischalifour: See 4b843ec6ba26b055ee4477af7f65aabd04590576.

packages/horizontal-slider-react/src/HorizontalSlider.tsx
shortcuts: Ooookay

packages/horizontal-slider-react/src/HorizontalSlider.tsx
francoischalifour: See 4b843ec6ba26b055ee4477af7f65aabd04590576.

shortcuts (approved):

francoischalifour:
Merging now to release on the `experimental` tag, but feel free to keep the discussion going about the translations and the `itemComponent` prop.

packages/horizontal-slider-react/src/HorizontalSlider.tsx
Haroenv: in https://reactnative.dev/docs/flatlist it's called `renderItem`, which isn't really better. What I made the comment initially about was that you should rename itemComponent to ItemComponent possibly? I'm not sure what the jsx rules are

PR #19, opened 2021-05-21
fix(recommendations): convert scores to integers

francoischalifour:
The Algolia engine is not able to return recommendations properly when the scores of the `optionalFilters` contain decimals. This makes sure to send integer scores to the engine.


shortcuts (approved):

Shipow (approved):

Haroenv (approved):
is it possible to have more than two numbers after the comma in the recommendations api? maybe round the number?

francoischalifour:
@Haroenv The current recommendations API only returns 2 decimals. The API will change soon though, and the specs are still to be defined, but we won't have to compute these `optionalFilters` anymore.

PR #22, opened 2021-05-25
feat(fbt): update filter strategy without fallbacks

francoischalifour:
This updates the _Frequently Bought Together_ (FBT) filter strategy for stricter matches.

## FBT without fallbacks

If FBT isn't provided any fallbacks, we want the strict recommended records coming from `record.recommendations`. We want strict ones because as opposed to Related Products, we cannot guess what products were bought with the reference products.

To get strict FBT, we now rely on the `facetFilters` search parameter.

## FBT with fallback

When FBT is provided fallbacks, we use the same strategy as Related Products. We compute `optionalFilters` based on the recommended records _and_ the fallback filters.

This strategy doesn't change in that PR.

---

Note that this filter strategy is implemented in the library for now, but it'll likely be part of the backend for GA.


Shipow (approved):

shortcuts (approved):
Looks good to me!

PR #23, opened 2021-05-25
feat(recommendations): add `fallbackComponent` prop

francoischalifour:
This introduces a new prop called `fallbackComponent` that is called when no recommendations are returned by the engine.

A pattern that we want to encourage is to display Related Products when there's no FBT, as seen in the demo.


shortcuts (approved):
Looks good!

Note for future improvements: we might want to pass some of the pre-defined parameters to the `fallbackComponent` function (I think `itemComponent`, `objectIDs` make sense)

PR #27, opened 2021-06-02
feat(recommendations): change renderer implementation to VDOM

francoischalifour:
This is a rework of the rendering system of the Recommendations components to an agnostic virtual DOM implementation.

*Note that the Horizontal Slider UI component is not part of this PR, it'll be migrated in another stacked PR.*

## Problems

Here's a list of the problems we have with the current architecture:

- **Bundle size**. The JavaScript package embeds the complete `preact/compat` package which exports the whole React API. (We now only rely on `preact` and `preact/hooks`.)
- **Not future-proof**. The current React/Preact aliasing is not going to allow us to bring support for Vue, React Native, and all VDOM-based framework.
- **Bridge between JavaScript packages and UI components**. The bridge between UI components like Horizontal Slider and the Recommendations component is challenging because we're not in control of the `view` prop rendering framework.
- **Tooling**. The tooling is difficult to get working because we have to alias the dependencies in all our toolchain (Parcel, Jest, Rollup, Babel).

### Current

The previous packages main entry point was `react-recommendations`, where the logic and the rendering happened. The `js-recommendations` package relied on `react-recommendations` with an alias from `react` to `preact/compat`.

### New

We now have these packages:

```
packages
├── recommendations-core
├── recommendations-js
├── recommendations-react
└── recommendations-vdom
```

### `recommendations-core`

`recommendations-core` is meant to be self-deprecated once we support Recommend in the API client. It embeds the logic for fetching the recommendations, using the Algolia Search API.

**API**

- `getRecommendations`
- `getFrequentlyBoughtTogether`
- `getRelatedProducts`

### `recommendations-vdom`

`recommendations-vdom` is an agnostic package that allows to build components given a renderer. A renderer is composed of `{ createElement, Fragment }`. It's purely the UI part and creates uncontrolled components, lacking any state and lifecycle functionalities.

**API**

- `createFrequentlyBoughtTogetherComponent`
- `createRelatedProductsComponent`

**Usage**

```tsx
import { createElement, Fragment } from 'react';

const FrequentlyBoughtTogether = createFrequentlyBoughtTogetherComponent({
  createElement,
  Fragment,
});
```

### `recommendations-js`

`recommendations-js` is an implementation of `recommendations-vdom` with Preact (similar front-end framework that we use in Autocomplete and InstantSearch). It specifies `recommendations-vdom` with the Preact renderer.

**API**

- `relatedProducts`
- `frequentlyBoughtTogether`

### `recommendations-react`

`recommendations-react` is an implementation of `recommendations-vdom` with React.

**API**

- `RelatedProducts`
- `FrequentlyBoughtTogether`
- `useRecommendations`
- `useRelatedProducts`
- `useFrequentlyBoughtTogether`

### Other packages in the future

When the time comes, we'll be able to create `recommendations-vue` with the same strategy as `recommendations-js` and `recommendations-react`. We could also expose these components for React Native in the future.

## How to review

I'd recommend reading from package to package, and we can do a live code review if you'd like!

## Next steps

- Deprecate the previous packages
- If we decide to move forward with this new architecture, we'll send a stacked PR to also migrate the HorizontalSlider UI components.


Haroenv:
Before reading the detailed code, I'm already letting know that I'm very excited about this layout of packages, it will solve a lot of problems we have now with mixing the preact/react in a single code base, making createElement & fragment opt-in is great (like in autocomplete) (in the mean time kinda thinking if it could make sense for InstantSearch, but I'm not really sure, since there the logic isn't just UI, but also clearly life cycle logic)

README.md
```diff
@@ -4,14 +4,14 @@
...
+> ⚠️ This branch is the experimental version of the UI components. **If you've been using the npm `beta` tag, make sure to [switch to the `experimental` tag](/packages/recommendations-react#installation)**.
 
 During the beta of Algolia Recommend, the UI components rely on an Algolia Index to retrieve the recommendations, and perform a regular Search API request to get the recommendations.
 
 ## Packages
 
-- [`@algolia/js-recommendations`](/packages/js-recommendations): JavaScript package for [Algolia Recommend](https://www.algolia.com/doc/guides/algolia-ai/recommend/)
-- [`@algolia/react-recommendations`](/packages/react-recommendations): React package for [Algolia Recommend](https://www.algolia.com/doc/guides/algolia-ai/recommend/)
+- [`@algolia/recommendations-js`](/packages/recommendations-js): JavaScript package for [Algolia Recommend](https://www.algolia.com/doc/guides/algolia-ai/recommend/)
+- [`@algolia/recommendations-react`](/packages/recommendations-react): React package for [Algolia Recommend](https://www.algolia.com/doc/guides/algolia-ai/recommend/)
```
Haroenv, line 14: does -core and -vdom need to be added to this list?

bundlesize.config.json
```diff
@@ -4,17 +4,17 @@
...
     },
     {
-      "path": "packages/react-recommendations/dist/umd/index.js",
-      "maxSize": "3 kB"
+      "path": "packages/recommendations-js/dist/umd/index.js",
+      "maxSize": "7.25 kB"
+    },
+    {
+      "path": "packages/recommendations-react/dist/umd/index.js",
+      "maxSize": "3.25 kB"
```
Haroenv, line 17: why did the react package increase in size? Is that the overhead of the createXXX functions?

examples/js-demo/package.json
```diff
@@ -8,17 +8,14 @@
     "build": "parcel build index.html",
     "start": "parcel index.html"
   },
-  "alias": {
-    "@algolia/js-recommendations": "@algolia/js-recommendations/dist/esm"
```
Haroenv, line 12: glad this isn't needed anymore

packages/recommendations-core/package.json
```diff
@@ -1,5 +1,5 @@
 {
-  "name": "@algolia/js-recommendations",
+  "name": "@algolia/recommendations-core",
```
Haroenv, line 2: wow this diff is totally from the wrong file

packages/recommendations-core/src/__tests__/getRecommendations.test.ts
```diff
@@ -57,12 +55,12 @@ function createRecommendationsClient() {
...
 }
 
-describe('useRecommendations', () => {
+describe('getRecommendations', () => {
   test('calls the correct index for "related-products"', async () => {
     const { index, searchClient } = createRecommendationsClient();
     const props = {
       model: 'related-products' as RecommendationModel,
-      searchClient,
+      searchClient: searchClient as any,
```
Haroenv, line 63: why is this needed? maybe a comment?

packages/recommendations-core/src/getFrequentlyBoughtTogether.ts
```diff
@@ -0,0 +1,21 @@
...
+  getRecommendations,
+  GetRecommendationsProps,
+} from './getRecommendations';
+
+export type GetFrequentlyBoughtTogetherProps<TObject> = Omit<
+  GetRecommendationsProps<TObject>,
+  'model' | 'fallbackFilters'
+>;
+
+export function getFrequentlyBoughtTogether<TObject>(
```
Haroenv, line 11: this name feels a little bit like an unfinished sentence, since it's already implying e-commerce with the word "bought", what about "getFrequentlyBoughtTogetherProducts"?

packages/recommendations-core/src/getRecommendations.ts
```diff
@@ -0,0 +1,178 @@
...
+          return {
+            recommendations: props.transformItems(hits),
+          };
+        });
+    })
+    .catch(() => {
+      // The `objectID` doesn't exist, we cannot get recommendations.
+      return {
+        recommendations: [],
+      };
```
Haroenv, line 176: should this not be an error instead of an empty state? It seems hard to debug

(sorry for commenting on unrelated code just because it renamed)

francoischalifour:
@Haroenv This architecture has also sparkled some ideas for InstantSearch in my mind.

For instance, we could rely on a single `HierarchicalMenu` VDOM UI component that InstantSearch (either JS, React or Vue) can reuse and pass props to it, just like we're doing in this PR for `RelatedProducts`: https://github.com/algolia/ui-components/blob/a2ac8d5bbb68c044d931e29b53b1ca832d174d83/packages/recommendations-react/src/RelatedProducts.tsx

This means that we would be able to use `HierarchicalMenu` in the context of InstantSearch, but also elsewhere, as long as the props have the right interface. It'd bring really interesting use cases for Query Categorization and other new features, outside of strictly searching, but rather browsing/discovering.

The `useRelatedProducts` state logic would be our connector logic. Thinking about it this way, all the state and lifecycle logic would be part of InstantSearch, which would then populate the UI components. (It'll make sense once you'll see @shortcuts stacked PR for the HorizontalSlider component.)

InstantSearch would become just the search state manager with an API idiomatic to each framework, and it'd rely on agnostic UI components, shared among all frameworks. Plugging the life cycle logic isn't an unknown anymore as long as those components remain uncontrolled (i.e., they don't use any frameworks' specifics like `useState`).

That's just one application that I thought about when try these new patterns, but we can definitely discuss it further!

packages/recommendations-js/README.md
```diff
@@ -1,13 +1,13 @@
...
 JavaScript package for [Algolia Recommend](https://www.algolia.com/doc/guides/algolia-ai/recommend/).
 
 ## Installation
 
 ```sh
-yarn add @algolia/js-recommendations@experimental
+yarn add @algolia/recommendations-js@experimental
 # or
-npm install @algolia/js-recommendations@experimental
+npm install @algolia/recommendations-js@experimental
```
Haroenv, line 10: we'll need to make sure to `npm deprecate` the older package names once this is merged

packages/recommendations-js/src/frequentlyBoughtTogether.tsx
```diff
@@ -0,0 +1,62 @@
...
+  {
+    createElement,
+    Fragment,
+  }
+);
+
+function useFrequentlyBoughtTogether<TObject>(
+  props: GetFrequentlyBoughtTogetherProps<TObject>
+) {
+  const [result, setResult] = useState<GetRecommendationsReturn<TObject>>({
```
Haroenv:
```suggestion
  const [result, setResult] = useState<GetRecommendationsResult<TObject>>({
```

I feel like this is named more naturally, since it's a network response/result, not really a return

packages/recommendations-react/src/useAlgoliaAgent.ts
```diff
@@ -0,0 +1,12 @@
+import { GetRelatedProductsProps } from '@algolia/recommendations-core';
+import { useEffect } from 'react';
+
+import { version } from './version';
+
+type UseAlgoliaAgentProps = Pick<GetRelatedProductsProps<any>, 'searchClient'>;
+
+export function useAlgoliaAgent(props: UseAlgoliaAgentProps) {
+  useEffect(() => {
```
Haroenv, line 9: in theory this code could also be shared between vdom implementations, but then you'll need to map to exactly useEffect in Vue, which might not be straightforward, so copied makes sense

packages/recommendations-vdom/README.md
```diff
@@ -0,0 +1,11 @@
+# `@algolia/recommendations-vdom`
+
+VDOM package for [Algolia Recommend](https://www.algolia.com/doc/guides/algolia-ai/recommend/).
```
Haroenv, line 3: I checked and react and preact both style it like this: VDOM, so it's good!

packages/recommendations-vdom/rollup.config.js
```diff
@@ -0,0 +1,27 @@
+import { plugins } from '../../rollup.base.config';
+
+import pkg from './package.json';
+
+if (!process.env.BUILD) {
+  throw new Error('The `BUILD` environment variable is required to build.');
+}
+
+const output = {
```
Haroenv, line 9: I wonder if the output could also be exported from the base rollup config (as a function that takes pkg), it's the same for all of them

packages/recommendations-vdom/src/DefaultFallback.tsx
```diff
@@ -0,0 +1,7 @@
+/** @jsx createElement */
```
Haroenv, line 1: it's a little confusing here because createElement isn't used as a prop. Should this maybe be inlined in where it's used, as it's not really a fallback, but nothing if it's not given

packages/recommendations-vdom/src/ListView.tsx
```diff
@@ -0,0 +1,40 @@
...
+          props.classNames.container
+        )}
+      >
+        <ol className={cx('auc-Recommendations-list', props.classNames.list)}>
+          {props.items.map((item) => (
+            <li
+              key={item.objectID}
+              className={cx('auc-Recommendations-item', props.classNames.item)}
+            >
+              <props.itemComponent item={item} />
```
Haroenv, line 33: theres' another prop called props.Header (capitalised), maybe it's nicer if all components that are props are pascal case? IIRC that's a requirement if it's not read from an object directly, so being consistent can help

packages/recommendations-vdom/src/RelatedProducts.tsx
```diff
@@ -0,0 +1,60 @@
...
+      <ViewComponent
+        classNames={classNames}
+        itemComponent={props.itemComponent}
+        items={props.items}
+        translations={translations}
+        {...viewProps}
+      />
+    );
+
+    return children({
```
Haroenv, line 51: should this use the children-as-a-function pattern still? not sure if that translates to Vue efficiently

Haroenv: if children is a component, all those arguments could be props, not sure what the other options are

packages/recommendations-vdom/src/types/ProductRecord.ts
```diff
@@ -0,0 +1,7 @@
+export type ProductRecord<TObject> = TObject & {
```
Haroenv: should this not be part of the -core package?

packages/recommendations-vdom/src/types/Renderer.ts
```diff
@@ -0,0 +1,36 @@
+export type Pragma = (
+  type: any,
+  props: Record<string, any> | null,
+  ...children: ComponentChildren[]
+) => JSX.Element;
```
Haroenv, line 5: where does JSX get defined?

packages/recommendations-vdom/src/types/Renderer.ts
```diff
@@ -0,0 +1,36 @@
...
+export type PragmaFrag = any;
+
+type ComponentChild =
+  | VNode<any>
```
Haroenv, line 9: and this?

ship.config.js
```diff
@@ -8,9 +8,9 @@ module.exports = {
     packagesToBump: [],
     packagesToPublish: [
       'packages/js-horizontal-slider',
```
Haroenv, line 10: these packages get migrated in a different PR once this is approved?

Haroenv (commented):
I love it!

ship.config.js
shortcuts: The migration is already done on an other branch: https://github.com/algolia/ui-components/tree/feat/vdom-pt2 but we will open the second PR once we get some feedbacks on this implementation :D

README.md
shortcuts: I was also wondering, I've added them to the migration to VDOM part 2 but should we only state the "real" packages we provide and not the internal ones?

packages/recommendations-core/src/__tests__/getRecommendations.test.ts
shortcuts: This might be because `createRecommendationsClient` creates a lite version of the `SearchClient` while `getRecommendations` doesn't?

packages/recommendations-vdom/rollup.config.js
shortcuts: Since we won't use rollup to build ESM I've inlined it in the part 2 of the migration, it's still different for the React packages since we use the `external` prop

packages/recommendations-vdom/src/types/Renderer.ts
shortcuts: https://github.com/algolia/ui-components/blob/a2ac8d5bbb68c044d931e29b53b1ca832d174d83/packages/recommendations-vdom/src/types/Renderer.ts#L18-L21

shortcuts (commented):
So good!

README.md
francoischalifour: I was hesitant because those packages will probably remain internal, but that's fair. Added in 10272d48acf39bcd214e4cb7bd564a2e14cdb717.

bundlesize.config.json
francoischalifour: Good question! It went from 3.0KB to 3.1KB. Looking at the diff, it's likely because Babel transpile code multiple times (because the VDOM package is a new dependency) and those polyfills end up being duplicated in the bundle.

I'm able to counter-balance the cost of the abstraction and get it back down to 3.02KB by using the `@babel/plugin-transform-runtime` plugin in e0f94b0006000527f20a227c3877cd55734b5762. Lmk what you think?

packages/recommendations-core/src/__tests__/getRecommendations.test.ts
francoischalifour: That's just temporary because the typings for the yet non-existant Recommendations API client don't exist. Since the expected search client and its typings are going to change very soon, I didn't bother hiding the typing issue.

packages/recommendations-core/src/getFrequentlyBoughtTogether.ts
francoischalifour: I agree—it would also be more consistent with "getRelated**Products**". The thing is it's been branded like that every where else on the AI team side. I'll catch up with them to see what they think about that.

packages/recommendations-core/src/getRecommendations.ts
francoischalifour: Since we mimic the final API with some workarounds with the Search endpoint, we didn't yet bother processing errors. I think we can rather tackle this more efficiently once we implement the new Recommendations API client spec, and have this discussion then. Wdyt?

packages/recommendations-js/src/frequentlyBoughtTogether.tsx
francoischalifour: Addressed in 38f20bc3b64e1dd097b7148a284ba3a6b201c293.

packages/recommendations-react/src/useAlgoliaAgent.ts
francoischalifour: Yeah, here we want to avoid to use framework specifics and share them across packages, otherwise we go back to the original building/aliasing issues easily. This "code duplication" will be counter-balance with some testing.

packages/recommendations-vdom/rollup.config.js
francoischalifour: We plan to reiterate on the build scripts after @shortcuts stacked PR! (it'll include a fully-shared Rollup config, banners with license, etc.)

packages/recommendations-vdom/src/DefaultFallback.tsx
francoischalifour: I see it rather consistent with other default components. For now the default fallback is to not display anything but we're not sure if it'll stay like that.

I'd say let's keep it like that for now if you don't mind too much?

packages/recommendations-vdom/src/ListView.tsx
francoischalifour: I agree we need to find some consistency. We also have `fallbackComponent`. I think it's out of scope for this PR but let's definitely address this before public release.

packages/recommendations-vdom/src/RelatedProducts.tsx
francoischalifour: Having children as a function makes the API powerful for custom use cases, but the fact that it might not work with Vue is a good point. We'll need to test it before public release then. I guess converting that to a slot would work in the Vue API?

packages/recommendations-vdom/src/types/ProductRecord.ts
francoischalifour: Indeed—will move them.

packages/recommendations-vdom/src/types/Renderer.ts
francoischalifour: It's a TypeScript global, we also use it in Autocomplete.

packages/recommendations-vdom/src/types/Renderer.ts
francoischalifour: `VNode` is defined below.

packages/recommendations-js/README.md
francoischalifour: I added that point in the "Next steps" section—thanks!

packages/recommendations-vdom/src/types/ProductRecord.ts
francoischalifour: See 2fba5ac75b74a6d4ad6b591e153a5f79f203bfed.

README.md
Haroenv: could be in a separate header for internal packages?

packages/recommendations-vdom/src/ListView.tsx
Haroenv: I'll let you open an issue / note it down somewhere and this can be resolved

packages/recommendations-vdom/src/RelatedProducts.tsx
Haroenv: most likely, I didn't think of slots, so that should indeed solve the issue

Haroenv (approved):
I think for now this is good, although the remaining comments probably should move to issues eventually

packages/recommendations-vdom/src/ListView.tsx
francoischalifour: [Created a ticket on JIRA](https://algolia.atlassian.net/browse/DX-2016).

packages/recommendations-react/src/useAlgoliaAgent.ts
```diff
@@ -0,0 +1,12 @@
...
+    props.searchClient.addAlgoliaAgent('recommendations-js', version);
```
sarahdayan:
```suggestion
    props.searchClient.addAlgoliaAgent('recommendations-react', version);
```

sarahdayan (commented):
Nice work, this layering feels a lot more future-proof. I'm also super excited to see how this can influence InstantSearch and maybe move the UI components outside, in this package 😍

@Haroenv and @shortcuts covered the PR and I agree with most comments, just added a minor one.

One question (unrelated to merging this PR)—with Autocomplete and now these UI components, it seems we're going with VDOM as the default for our vanilla JavaScript implementations. **Is it the direction we want to take? Are we also going to this route for InstantSearch.js?**

Personally I think it makes a lot of sense for performance and convenience concerns, and templating solutions don't come even close to what JSX expressions bring to the table. But this is also an opinionated choice which "force" users to use a VDOM implem and write JSX (or build VDOM trees by hand). I could see some users confused by this choice and I'm wondering if this is something we've vetted beforehand.

Haroenv:
> Personally I think it makes a lot of sense for performance and convenience concerns, and templating solutions don't come even close to what JSX expressions bring to the table. But this is also an opinionated choice which "force" users to use a VDOM implem and write JSX (or build VDOM trees by hand). I could see some users confused by this choice and I'm wondering if this is something we've vetted beforehand.

I agree with your Sarah, For these smaller packages I think this makes sense, but I'm not fully convinced whether that's the right solutions for InstantSearch.js, templates (while having lots of problems) are a lot more approachable for a dev who's never used build systems or jsx

sarahdayan:
@Haroenv Yeah, that's debatable though. When using a "small" package you might expect proportionate efforts, with less setup and a more plug-and-play approach.

There's also the concern of consistency, where developers using one of our UI solutions then move on to the next and there's a different approach to templating. Or even use them altogether and there are two ways of doing things.

francoischalifour:
I've looked around quite a lot and for now I think that VDOM is the choice that **makes it easy to build interactive UIs** (which is the main point when dealing with search and discovery) both **for us as maintainers and for devs using our libs**.

(In InstantSearch, we rely on Preact internally, which uses VDOM, and the user-facing templating system is HTML/Hogan.js.)

I want to stress that what's interesting with this VDOM approach is that we can **offer users to easily bail out of JSX**, while still offering this VDOM layer. For example, we could detect that users are returning HTML strings and wrap that in a `createElement('div', { dangerouslySetInnerHTML: { __html: content }})`. Users could therefore **use Hogan.js if they need to avoid XSS**.

Example:

```tsx
relatedProducts({
  // ...
  itemComponent({ item }) {
    // We're able to detect HTML and forward it to `dangerouslySetInnerHTML`
    return `<h3>${item.name}</h3>`;
  },
});
```

What I'm trying to say is that **this abstraction is still permissive**, and it still allows users to use HTML if they want. If they express this need, we can create this layer to facilitate this alternative.

I also express some problems in solutions that don't provide diffing in the original Autocomplete VDOM PR: algolia/autocomplete#381. Accessibility is another concern that we already suffer from in InstantSearch.js.

As always, I'm open to any alternatives you've thought about.

packages/recommendations-react/src/useAlgoliaAgent.ts
francoischalifour: See 3c6e528ff1e233eb091b529c6386c15be789c0da.

shortcuts (approved):
🚀

sarahdayan:
@francoischalifour Again, I'm fully sold on the benefits of using VDOM internally, and believe this is a sane pick for users as well. I think it's awesome if we can facilitate usage of template strings or a third-party template engine, and forward that ourselves to the `dangerouslySetInnerHTML` API because it caters to more users and avoid alienating those who have never seen JSX before.

Ultimately, users don't need to know we're using VDOM under the hood (just like with InstantSearch.js), it should be about what they want to use, and for this to feel natural.

francoischalifour:
@sarahdayan I added "[Detect and forward HTML templates to `dangerouslySetInnerHTML`](https://algolia.atlassian.net/browse/DX-2085?atlOrigin=eyJpIjoiNTZhN2JlMjI4YTllNGYyMmI4ZTU3NTQ5MGJlYmRjNDciLCJwIjoiaiJ9)" to the backlog. It'll also be a nice addition to Autocomplete.

PR #28, opened 2021-06-03
feat(horizontal-slider): change renderer implementation to VDOM

shortcuts:
_Follow up of the previous PR: https://github.com/algolia/ui-components/pull/27_

This is a rework of the rendering system of the Horizontal Slider UI components to an agnostic virtual DOM implementation.

## Problems

Here's a list of the problems we have with the current architecture:

- **Bundle size**. The JavaScript package embeds the complete `preact/compat` package which exports the whole React API. (We now only rely on `preact` and `preact/hooks`.)
- **Not future-proof**. The current React/Preact aliasing is not going to allow us to bring support for Vue, React Native, and all VDOM-based framework.
- **Bridge between JavaScript packages and UI components**. The bridge between UI components like Horizontal Slider and the Recommendations component is challenging because we're not in control of the `view` prop rendering framework.
- **Tooling**. The tooling is difficult to get working because we have to alias the dependencies in all our toolchain (Parcel, Jest, Rollup, Babel).

### Current

The previous packages main entry point was `react-horizontal-slider`, where the logic and the rendering happened. The `js-horizontal-slider` package relied on `react-horizontal-slider` with an alias from `react` to `preact/compat`.

### New

We now have these packages:

```
packages
├── horizontal-slider-js
├── horizontal-slider-react
├── horizontal-slider-theme
└── horizontal-slider-vdom
```

### `horizontal-slider-js`

`horizontal-slider-js` is an implementation of `horizontal-slider-vdom` with Preact (similar front-end framework that we use in Autocomplete, InstantSearch and the recommendations-js package). It specifies `horizontal-slider-vdom` with the Preact renderer.

**API**

- `horizontalSlider`

### `horizontal-slider-react`

`horizontal-slider-react` is an implementation of `horizontal-slider-vdom` with React.

**API**

- `HorizontalSlider`

### `horizontal-slider-theme`

Package that hosts the `HorizontalSlider` component default theme.

### `horizontal-slider-vdom`

`horizontal-slider-vdom` is an agnostic package that allows to build components given a renderer. A renderer is composed of `{ createElement, Fragment }`. It's purely the UI part and creates uncontrolled components, lacking any state and lifecycle functionalities.

**API**

- `createHorizontalSliderComponent`
- `generateHorizontalSliderId`: the function to initialize the `sliderIdRef` ref with.
- `updateNavigationButtonsProps`: the function to wrap in an effect, which will update the current value of the refs.

### `container`

The `container` props for the `recommendations-js` and `horizontal-slider-js` packages are now optional. When `undefined`, we will directly return the component instead of rendering it in its container.

This allows us to:
- Leverage the `view` prop of the `recommendations-js` components with the `HorizontaSlider`.
- Render a `fallbackComponent` from the `recommendations-js` library in the same container as its parent.

**Usage**
```js
frequentlyBoughtTogether({
  container: '#frequentlyBoughtTogether',
  searchClient,
  indexName,
  // ...
  fallbackComponent() {
    return relatedProducts({
      searchClient,
      indexName,
      view: horizontalSlider,
      // ...
    });
  },
});
```

### Other packages in the future

When the time comes, we'll be able to create `horizontal-slider-vue` with the same strategy as `horizontal-slider-js` and `horizontal-slider-react`. We could also expose this component for React Native in the future.

## Next steps

- Deprecate the previous packages
- Update rollup bundle config


examples/js-demo/index.html
```diff
@@ -16,6 +16,7 @@ <h1>Algolia Recommendations</h1>
 
       <div id="autocomplete"></div>
 
+      <div id="slider"></div>
```
francoischalifour: We don't need this container in the demo.

packages/horizontal-slider-js/README.md
```diff
@@ -43,7 +43,7 @@ horizontalSlider({
 
 ### `container`
 
-> `string | HTMLElement` | **required**
+> `string | HTMLElement` |
```
francoischalifour:
```suggestion
> `string | HTMLElement`
```

packages/horizontal-slider-vdom/src/index.ts
```diff
@@ -0,0 +1,2 @@
+export * from './types';
+export * from './HorizontalSlider';
```
francoischalifour: We can sort those ascending like in other files.

packages/horizontal-slider-vdom/src/types/FrameworkProps.ts
```diff
@@ -0,0 +1,11 @@
+type Ref<TElement> = {
+  current: TElement;
+};
```
francoischalifour, line 3: I'm wondering how that will play out with Vue because [their refs are strings](https://v3.vuejs.org/guide/component-template-refs.html).

packages/horizontal-slider-vdom/src/types/Renderer.ts
```diff
@@ -0,0 +1,36 @@
+export type Pragma = (
```
francoischalifour, line 1: Next component that we create we might want have a shared/common package for these types.

packages/recommendations-js/README.md
```diff
@@ -60,7 +97,7 @@ The function accepts all the [shared props](#shared-props) and the following:
 
 ##### `container`
 
-> `string | HTMLElement` | **required**
+> `string | HTMLElement` |
```
francoischalifour:
```suggestion
> `string | HTMLElement`
```

packages/recommendations-js/README.md
```diff
@@ -198,7 +235,7 @@ The function accepts all the [shared props](#shared-props) and the following:
...
```
francoischalifour:
```suggestion
> `string | HTMLElement`
```

francoischalifour (approved):
Looks good—love the optional `container` pattern!

I've got a concern about the ref compatibility with a future Vue package though.

packages/horizontal-slider-vdom/src/types/Renderer.ts
shortcuts: Yep, I can already see multiple use cases for a shared package!

packages/recommendations-js/README.md
shortcuts: Fixed in: https://github.com/algolia/ui-components/pull/28/commits/0aef9188e2877989fb4c919766a0265db64f24e4

packages/recommendations-js/README.md
shortcuts: Fixed in: https://github.com/algolia/ui-components/pull/28/commits/0aef9188e2877989fb4c919766a0265db64f24e4

packages/horizontal-slider-js/README.md
shortcuts: Fixed in: https://github.com/algolia/ui-components/pull/28/commits/0aef9188e2877989fb4c919766a0265db64f24e4

examples/js-demo/index.html
shortcuts: Oops I forgot this one: fixed in https://github.com/algolia/ui-components/pull/28/commits/3e49a8bf3105aba0cca8baa4d96b5741d49df393

packages/horizontal-slider-vdom/src/index.ts
shortcuts: Fixed in: https://github.com/algolia/ui-components/pull/28/commits/e6852336f417ec12d89d5931960c83d2c7efcda0

packages/horizontal-slider-vdom/src/types/FrameworkProps.ts
shortcuts: Since we will be in control of the Vue implementation, we could define a fallback type for it

README.md
```diff
@@ -10,12 +10,14 @@ During the beta of Algolia Recommend, the UI components rely on an Algolia Index
...
-- [`@algolia/recommendations-core`](/packages/recommendations-react): Core package for [Algolia Recommend](https://www.algolia.com/doc/guides/algolia-ai/recommend/)
 - [`@algolia/recommendations-vdom`](/packages/recommendations-react): VDOM package for [Algolia Recommend](https://www.algolia.com/doc/guides/algolia-ai/recommend/)
-- [`@algolia/ui-components-js-horizontal-slider`](/packages/js-horizontal-slider): JavaScript horizontal slider UI component
-- [`@algolia/ui-components-react-horizontal-slider`](/packages/react-horizontal-slider): React horizontal slider UI component
+- [`@algolia/ui-components-horizontal-slider-js`](/packages/horizontal-slider-js): JavaScript horizontal slider UI component
+- [`@algolia/ui-components-horizontal-slider-react`](/packages/horizontal-slider-react): React horizontal slider UI component
+- [`@algolia/ui-components-horizontal-slider-theme`](/packages/horizontal-slider-theme): Horizontal slider UI component theme
+- [`@algolia/ui-components-horizontal-slider--vdom`](/packages/horizontal-slider-react): VDOM package for Horizontal slider UI component
```
Haroenv:
```suggestion
- [`@algolia/ui-components-horizontal-slider-vdom`](/packages/horizontal-slider-vdom): VDOM package for Horizontal slider UI component
```

package.json
```diff
@@ -42,10 +42,10 @@
     "@types/react": "^17.0.5",
     "@typescript-eslint/eslint-plugin": "2.19.0",
     "@typescript-eslint/parser": "2.19.0",
+    "autoprefixer": "10.0.4",
     "babel-eslint": "10.1.0",
-    "babel-plugin-module-resolver": "4.1.0",
     "bundlesize": "0.18.0",
-    "cross-env": "7.0.2",
```
Haroenv, line 48: are these two removals on purpose?

packages/horizontal-slider-js/README.md
```diff
@@ -1,13 +1,13 @@
-# `@algolia/ui-components-js-horizontal-slider`
+# `@algolia/ui-components-horizontal-slider-js`
```
Haroenv, line 1: much better name, alternative would have been `js-ui-components- horizontal-slider`, but this format works well

packages/horizontal-slider-js/src/horizontalSlider.tsx
```diff
@@ -0,0 +1,68 @@
...
+  const previousButtonRef = useRef<HTMLButtonElement>(null);
+  const sliderIdRef = useRef(generateHorizontalSliderId());
+
+  useEffect(() => {
+    updateNavigationButtonsProps({
+      listRef,
+      nextButtonRef,
+      previousButtonRef,
+    });
+  });
```
Haroenv, line 34: should this happen on every change?

packages/horizontal-slider-js/src/horizontalSlider.tsx
```diff
@@ -0,0 +1,68 @@
...
+    />
+  );
+}
+
+export function horizontalSlider<TObject extends RecordWithObjectID>({
+  container,
+  environment = window,
+  ...props
+}: HorizontalSliderProps<TObject> & EnvironmentProps) {
+  const children = <HorizontalSlider {...props} />;
```
Haroenv, line 59: should this be called children or component (it's only one element, maybe child)?

packages/horizontal-slider-react/src/HorizontalSlider.tsx
```diff
@@ -0,0 +1,43 @@
...
+    updateNavigationButtonsProps({ listRef, nextButtonRef, previousButtonRef });
+  });
+
+  return (
+    <UncontrolledHorizontalSlider
+      {...props}
+      listRef={listRef}
+      nextButtonRef={nextButtonRef}
+      previousButtonRef={previousButtonRef}
+      sliderIdRef={sliderIdRef}
```
Haroenv, line 33: I wonder if this should be a single refs object?

Haroenv: probably not due to the diff annoyances that causes

packages/horizontal-slider-vdom/src/HorizontalSlider.tsx
```diff
@@ -0,0 +1,163 @@
...
+  HorizontalSliderProps,
+  HorizontalSliderTranslations,
+  RecordWithObjectID,
+  Renderer,
+} from './types';
+
+let lastHorizontalSliderId = 0;
+
+export function generateHorizontalSliderId() {
+  return `uic-horizontal-slider-${lastHorizontalSliderId++}`;
```
Haroenv, line 13: does this generate stable ids for use in server side rendering?

packages/recommendations-js/README.md
```diff
@@ -26,12 +26,49 @@ To get started, you need a container for your results to go in. If you don’t h
...
+relatedProducts({
+  container: '#relatedProducts',
+  searchClient,
+  indexName,
+  objectIDs: [currentObjectID],
+  itemComponent({ item }) {
+    return (
+      <pre>
+        <code>{JSON.stringify(item)}</code>
+      </pre>
```
Haroenv, line 53: can you use it without preact (document.createElement) too, or does that not make sense?

package.json
Haroenv: seems like it! congrats

packages/horizontal-slider-react/src/HorizontalSlider.tsx
shortcuts: I feel like this is easier to understand as single refs, what would your implementation looks like?

packages/horizontal-slider-js/src/horizontalSlider.tsx
shortcuts: We choosed `children` to keep it consistent with the return of our `recommendations` components

packages/horizontal-slider-vdom/src/HorizontalSlider.tsx
shortcuts: This is the same logic as we used in [autocomplete](https://github.com/algolia/autocomplete) but I didn't tried it yet with SSR!

francoischalifour: For this to be SSR-safe, we need to accept an `id` prop in all our components. You would need to generate the `id` on the server and pass it to the client. (Autocomplete allows this.)

We've planed to add this later. But as it is now, it's not SSR-safe.

(Other interesting resources could be the new React's [`useOpaqueIdentifier`](https://github.com/facebook/react/pull/17322).)

packages/horizontal-slider-react/src/HorizontalSlider.tsx
Haroenv: `refs={{ nextButton: nextButtonRef, previousButton: previousButtonRef, sliderId: sliderIdRef }}` or something similar

packages/horizontal-slider-js/src/horizontalSlider.tsx
francoischalifour: `component` also makes sense to me. Both are fine IMO.

packages/recommendations-js/README.md
francoischalifour: That would be either VDOM or HTML strings (see https://github.com/algolia/ui-components/pull/27#issuecomment-856578530).

packages/horizontal-slider-js/src/horizontalSlider.tsx
shortcuts: Since this call will set update the `hidden` boolean to hide/show our arrows, yes. Did you meant to add the concerned refs to the dependencies?

francoischalifour (approved):

Haroenv (approved):
all good for me

PR #31, opened 2021-06-08
chore(build): unify Rollup configuration

francoischalifour:
This uses a single Rollup configuration for all our UMD builds.

## React UMD build

When `BUILD=react`, we provide React-specific Rollup setups.

## Banner

This update also includes a banner for our UMD build (that slightly increases the bundle size).

When built locally:

```
/*! @algolia/ui-components-horizontal-slider-react 1.0.0-experimental.2 (UNRELEASED e297a92) | MIT License | © Algolia, Inc. and contributors | https://github.com/algolia/ui-components */
```

When built for release:

```
/*! @algolia/ui-components-horizontal-slider-react 1.0.0-experimental.2 | MIT License | © Algolia, Inc. and contributors | https://github.com/algolia/ui-components */
```

### DIFF
--- a/scripts/rollup/config.js
+++ b/scripts/rollup/config.js
@@ -8,8 +8,6 @@ import { terser } from 'rollup-plugin-terser';
 
 import { getBundleBanner } from '../getBundleBanner.mjs';
 
-const isReactBuild = process.env.BUILD === 'react';
-
 const extensions = ['.js', '.jsx', '.ts', '.tsx', '.json'];
 const plugins = [
   replace({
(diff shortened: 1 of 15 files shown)


shortcuts (approved):
Looks clean!

scripts/rollup/config.js
```diff
@@ -30,3 +35,23 @@ export const plugins = [
     showGzippedSize: true,
   }),
 ];
+const external = isReactBuild ? ['react', 'react-dom'] : undefined;
+const globals = isReactBuild
+  ? { react: 'React', 'react-dom': 'ReactDOM' }
+  : undefined;
```
Haroenv: could this not be an argument to createRollupConfig?

Haroenv (commented):
makes sense, but I feel like the cross-env could be avoided

scripts/rollup/config.js
francoischalifour: It could, both are fine imo. Do you see any drawbacks with this?

Haroenv: I think an argument makes more sense, since you don't want to toggle between react and non-react builds for the same package, but either way works indeed

Haroenv (approved):
not blocking, I leave it up to you to decide if it's worth migrating or not

scripts/rollup/config.js
francoischalifour: I use another function called `createRollupConfigForReact` so we don't depend on env variables in 2c5389228058fe014449b3b963eb940539f9bbce.

Haroenv (approved):

PR #33, opened 2021-06-10
feat: rename packages to Recommend

francoischalifour:
Long term, we want to have all Algolia Recommend-related APIs under these packages, so this PR renames all our modules from `@algolia/recommendations-*` to `@algolia/recommend-*`.

This decision was approved by @mthuret.

(Note that the JS `getRecommendations` and React `useRecommendations` APIs weren't renamed.)


Haroenv (approved):
looks like this is correct 👍

packages/recommend-react/README.md
```diff
@@ -254,12 +254,9 @@ function App({ currentObjectID }) {
   return (
     <div className="auc-Recommendations">
```
shortcuts:
```suggestion
    <div className="auc-Recommend">
```

(there's other occurrences of this class)

shortcuts (approved):
https://github.com/algolia/ui-components/blob/ea7c08aa99fcb9ff6eb8bae009f315ccabedd8f5/packages/recommend-vdom/src/DefaultChildren.tsx#L13

I think this class should also be `auc-Recommend`, otherwise everything seems good!

packages/recommend-react/README.md
francoischalifour: Good catch—fixed in 3c562450fc18f8f6628385764b4d6f1dddb0b0ad.

shortcuts (approved):

PR #35, opened 2021-06-15
fix(react): use FBT hook in `FrequentlyBoughtTogether` component

dfcook:
React component for "Frequently Bought Together" recommendations was importing the `useRelatedProducts` hook so having both components on the same page resulted in duplicate results being displayed


francoischalifour (approved):
Oops—thank you!

PR #37, opened 2021-06-16
fix(recommend): ignore `items` type in components

francoischalifour:
This fixes a typing issue raised by the Dashboard team where `items` was marked as required in our FBT and RP components.

We forgot to omit the `items` prop which is only used by the VDOM components.

I took the opportunity to fix other typing inconsistencies where we expected the VDOM component to receive `GetRelatedProductsProps`, whereas it's only expect at the renderer packages' level (React and JS).

### DIFF
--- a/packages/recommend-react/src/RelatedProducts.tsx
+++ b/packages/recommend-react/src/RelatedProducts.tsx
@@ -1,6 +1,7 @@
+import { GetRelatedProductsProps } from '@algolia/recommend-core';
 import {
   createRelatedProductsComponent,
-  RelatedProductsProps,
+  RelatedProductsProps as RelatedProductsVDomProps,
 } from '@algolia/recommend-vdom';
 import React, { createElement, Fragment } from 'react';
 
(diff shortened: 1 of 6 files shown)


packages/recommend-react/src/RelatedProducts.tsx
```diff
@@ -1,6 +1,7 @@
+import { GetRelatedProductsProps } from '@algolia/recommend-core';
 import {
   createRelatedProductsComponent,
-  RelatedProductsProps,
+  RelatedProductsProps as RelatedProductsVDomProps,
```
shortcuts:
```suggestion
  RelatedProductsProps as RelatedProductsVDOMProps,
```

Since it is the official casing and we capitalize it in the repo, I think it make sense to also do it for the types

francoischalifour: See e95320836fcda0206607e213f1a85a1685e684f9.

Haroenv (approved):
good call!

shortcuts (approved):

PR #38, opened 2021-06-17
fix(fallbackFilters): rename `fallbackFilters` to `fallbackParameters`

shortcuts:
**Summary**

`fallbackFilters` will be named `fallbackParameters`  after the GA release, its structure also changed since it will accept `searchParameters`, but we only handle `facetFilters` for now.

### DIFF
--- a/packages/recommend-core/src/getRecommendations.ts
+++ b/packages/recommend-core/src/getRecommendations.ts
@@ -37,7 +37,7 @@ export type GetRecommendationsProps<TObject> = {
 
 export type GetRecommendationsInternalProps<TObject> = Required<
   GetRecommendationsProps<TObject>
->;
+> & { fallbackParameters: Required<Pick<SearchOptions, 'facetFilters'>> };
 
 export type GetRecommendationsResult<TObject> = {
   recommendations: Array<RecordWithObjectID<TObject>>;
@@ -47,11 +47,13 @@ function getDefaultedProps<TObject>(
   props: GetRecommendationsProps<TObject>
 ): GetRecommendationsInternalProps<TObject> {
   return {
-    fallbackParameters: { facetFilters: [] },
     maxRecommendations: 0,
     threshold: 0,
     transformItems: (items) => items,
     ...props,
+    fallbackParameters: {
+      facetFilters: props.fallbackParameters?.facetFilters || [],
+    },
     searchParameters: {
       analytics: false,
       analyticsTags: [`alg-recommend_${props.model}`],
--- a/packages/recommend-core/src/getRecommendations.ts
+++ b/packages/recommend-core/src/getRecommendations.ts
@@ -20,24 +20,34 @@ type RecommendedRecord = RecordWithObjectID<{
   recommendations?: RecommendRecord[];
 }>;
 
+export type RecommendSearchParameters = Omit<
+  SearchOptions,
+  'page' | 'hitsPerPage' | 'offset' | 'length'
+>;
+
 export type GetRecommendationsProps<TObject> = {
   model: RecommendModel;
   indexName: string;
   objectIDs: string[];
   searchClient: SearchClient;
 
-  fallbackParameters?: Pick<SearchOptions, 'facetFilters'>;
+  fallbackParameters?: RecommendSearchParameters;
   maxRecommendations?: number;
-  searchParameters?: SearchOptions;
+  searchParameters?: RecommendSearchParameters;
   threshold?: number;
   transformItems?: (
     items: Array<ProductRecord<TObject>>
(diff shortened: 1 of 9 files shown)


packages/recommend-core/src/getRecommendations.ts
```diff
@@ -26,7 +26,7 @@ export type GetRecommendationsProps<TObject> = {
   objectIDs: string[];
   searchClient: SearchClient;
 
-  fallbackFilters?: SearchOptions['optionalFilters'];
+  fallbackParameters?: { facetFilters: SearchOptions['facetFilters'] };
```
francoischalifour:
```suggestion
  fallbackParameters?: Pick<SearchOptions, 'facetFilters'>;
```

packages/recommend-core/src/getRecommendations.ts
```diff
@@ -47,7 +47,7 @@ function getDefaultedProps<TObject>(
   props: GetRecommendationsProps<TObject>
 ): GetRecommendationsInternalProps<TObject> {
   return {
-    fallbackFilters: [],
+    fallbackParameters: { facetFilters: [] },
```
francoischalifour: We either need to deeply default it (because if user provides `facetFilters: {}`, the current flow would throw), or handle code path expecting it to also be `undefined`. I'm fine with either way since this code will be thrown next week.

```suggestion
    fallbackParameters: {},
```

packages/recommend-core/src/utils/getHitsPerPage.ts
```diff
@@ -1,17 +1,17 @@
...
 };
 
 export function getHitsPerPage<TObject>({
-  fallbackFilters,
+  fallbackParameters,
   maxRecommendations,
   recommendationsCount,
 }: GetHitsPerPageParams<TObject>) {
-  const hasFallback = fallbackFilters.length > 0;
+  const hasFallback = fallbackParameters.facetFilters!.length > 0;
```
francoischalifour: We don't necessarily have `fallbackParameters.facetFilters` anymore so this line can throw.

```suggestion
  const hasFallback = Array.isArray(fallbackParameters.facetFilters) && fallbackParameters.facetFilters.length > 0;
```

packages/recommend-core/src/utils/getSearchParametersForModel.ts
```diff
@@ -4,47 +4,47 @@ import { GetRecommendationsInternalProps } from '../getRecommendations';
...
-    return fallbackFilters;
+    return fallbackParameters.facetFilters;
   }
 
   const recommendationFilters = recommendations
     .filter((recommendation) => recommendation.score > threshold)
     .map(({ objectID, score }) => `objectID:${objectID}<score=${score * 100}>`);
 
-  return [...recommendationFilters, ...fallbackFilters];
+  return [...recommendationFilters, ...fallbackParameters.facetFilters!];
```
francoischalifour: This could also probably throw if `facetFilters` is not defaulted.

```suggestion
  const facetFilters = fallbackParameters.facetFilters || []

  if (recommendations.length === 0) {
    return facetFilters;
  }

  const recommendationFilters = recommendations
    .filter((recommendation) => recommendation.score > threshold)
    .map(({ objectID, score }) => `objectID:${objectID}<score=${score * 100}>`);

  return [...recommendationFilters, ...facetFilters];
```

packages/recommend-core/src/utils/getSearchParametersForModel.ts
```diff
@@ -4,47 +4,47 @@ import { GetRecommendationsInternalProps } from '../getRecommendations';
...
 }
 
 function getSearchParametersForFrequentlyBoughtTogether<TObject>({
-  fallbackFilters,
+  fallbackParameters,
   recommendations,
   threshold,
 }: GetSearchParametersParams<TObject>): SearchOptions {
-  if (fallbackFilters.length === 0) {
+  if (fallbackParameters.facetFilters!.length === 0) {
```
francoischalifour: We need to check that it's an array first.

packages/recommend-core/src/getRecommendations.ts
shortcuts: There was some small issues when defaulting it to an empty object, does the other way around works to you? https://github.com/algolia/ui-components/pull/38/commits/496799dbe425f9eafbbf513703d6c369a9165d97

I thought it would be easier than adding conditions

shortcuts:
Closing this PR since most changes will be deprecated after the release of the new Recommend API client: https://github.com/algolia/algoliasearch-client-javascript/pull/1280

Update will come in a future PR.

PR #42, opened 2021-06-28
fix(recommend): only sort items with scores

francoischalifour:
This fixes cases where non-scored items are re-sorted because of our `sortBy` logic:

image

Notice that the items with `undefined` as score were re-sorted.

I reversed the `sortBy` predicate, and now only sort items if their score are different, otherwise keep the order returned by the engine.

### DIFF
--- a/packages/recommend-core/src/utils/__tests__/mapToRecommendations.test.ts
+++ b/packages/recommend-core/src/utils/__tests__/mapToRecommendations.test.ts
@@ -1,0 +1,129 @@
+import { mapToRecommendations } from '../mapToRecommendations';
+
+const response = {
+  results: [
+    {
+      exhaustiveNbHits: true,
+      hits: [
+        {
+          objectID: '1-1',
+          name: 'Product 1-1',
+          _score: 100,
+        },
+        {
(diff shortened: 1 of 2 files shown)


packages/recommend-core/src/utils/__tests__/mapToRecommendations.test.ts
```diff
@@ -0,0 +1,129 @@
...
+      hitsPerPage: 10,
+      nbHits: 10,
+      nbPages: 1,
+      page: 0,
+      processingTimeMS: 1,
+    },
+  ],
+};
+
+describe('mapToRecommendations', () => {
```
shortcuts, line 74: We can add a test to ensure that it doesn't sort a well-sorted response

shortcuts (approved):
Small comment but looks good to me!

packages/recommend-core/src/utils/__tests__/mapToRecommendations.test.ts
francoischalifour: As discussed in DM, the test already covers this case.

PR #43, opened 2021-06-28
feat(recommend): add status state (loading, stalled, idle)

francoischalifour:
## The problem

There was an issue where we rendered the `fallbackComponent` during first render because we checked `props.recommendations.length === 0`, which is the case before sending the request. Thus, when using `RelatedProducts` in `fallbackComponent`, it sent another request, which was unnecessary.

## This solution

This PR adds the loading status of the component in the API, and uses it to render the fallback component when there's no recommendations _and_ that the status is `idle`.

Possible statuses are `idle`, `loading` and `stalled`. The stalled status is useful to render product placeholders after `x`ms, to reduce UI flashes as opposed to rendering it when `loading`.

In the future, we might want to expose the stalled threshold value to the user, so that they can synchronize their placeholder content with other pieces on their website.

## Considerations

Notice that the `useStatus` hook in Preact and in React have an identical implementation, which is a bummer. I don't think we can use the same code here without additional tooling.

### DIFF
--- a/packages/recommend-react/src/useStatus.ts
+++ b/packages/recommend-react/src/useStatus.ts
@@ -1,0 +1,21 @@
+import { RecommendStatus } from '@algolia/recommend-vdom';
+import { useEffect, useRef, useState } from 'react';
+
+export function useStatus(initialStatus: RecommendStatus) {
+  const timerId = useRef<ReturnType<typeof setTimeout> | undefined>(undefined);
+  const [status, setStatus] = useState<RecommendStatus>(initialStatus);
+
+  useEffect(() => {
+    if (status !== 'stalled' && timerId.current) {
+      clearTimeout(timerId.current);
+    }
+
+    if (status === 'loading') {
+      timerId.current = setTimeout(() => {
+        setStatus('stalled');
+      }, 300);
+    }
+  }, [status]);
+
+  return { status, setStatus };
+}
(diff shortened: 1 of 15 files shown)


packages/recommend-react/src/useStatus.ts
```diff
@@ -0,0 +1,23 @@
+import { RecommendStatus } from '@algolia/recommend-vdom';
+import { useEffect, useRef, useState } from 'react';
```
Haroenv, line 2: could we make a factory where useEffect, useRef, useState as passed to the hook creator, so we can share the implementation?

Haroenv (commented):
Code looks good, but I think we could have a useStatus factory to share between react & preact

packages/recommend-react/src/useStatus.ts
shortcuts: Where would you see this implementation? If we have this duplication more frequently in the future, we could have a `recommend-hooks` package but right now it doesn't fit in `VDOM` or `core` IMO

Haroenv: it could fit in vdom, seeing as only react-like api lives there at the moment. Whenever we would add Vue, I guess a different api than purely hooks would be needed, but not yet atm

francoischalifour: I think it's a bit too soon to abstract this, but we should definitely define a strategy if these duplications need to grow.

shortcuts (approved):
Looks clean!

Haroenv (approved):
let's keep it in mind for the future

PR #46, opened 2021-06-29
fix(release): bump packages' versions on release

francoischalifour:
The `core` and `vdom` packages' versions were not bumped by Ship.js on release.


shortcuts (approved):
Good catch

PR #51, opened 2021-07-20
fix(core): avoid sending UI props to `recommend` client

shortcuts:
**Summary**

We now destructure the props before sending them to the client.


francoischalifour (commented):
Nice! Perhaps we can [add a test in `core`](https://github.com/algolia/recommend/blob/3f490fa951072e173b8b18c71e154adeffe82e76/packages/recommend-core/src/__tests__/getRecommendations.test.ts) to make sure that additional props are not forwarded to the client?

shortcuts:
> Nice! Perhaps we can [add a test in `core`](https://github.com/algolia/recommend/blob/3f490fa951072e173b8b18c71e154adeffe82e76/packages/recommend-core/src/__tests__/getRecommendations.test.ts) to make sure that additional props are not forwarded to the client?

Sure, added in https://github.com/algolia/recommend/pull/51/commits/353c1a9cd672102dc8c9414cf153f0966ee77824

packages/recommend-core/src/__tests__/getFrequentlyBoughtTogether.test.ts
```diff
@@ -0,0 +1,26 @@
+import { createRecommendationsClient } from '../../../../test/utils';
```
francoischalifour: `createRecommendationsClient` shouldn't be necessary now that we have `createRecommendClient`.

shortcuts: Make sense, is that what you meant? https://github.com/algolia/recommend/pull/51/commits/5c06a666e9f4b3c88372a801956c0bf55f3bd32f

francoischalifour (approved):

packages/recommend-core/src/__tests__/getFrequentlyBoughtTogether.test.ts
francoischalifour: Yep!

PR #52, opened 2021-07-30
fix(react): prevent infinite loops in exposed hooks

### LINKED ISSUE #49: useRecommendations & useRelatedProducts & useFrequentlyBoughtTogether cause infinite loop due to useEffect
## Description
`useRecommendations` & `useRelatedProducts` & `useFrequentlyBoughtTogether` hooks are causing the app to infinitely loop due to the useEffect dependencies. 

## Reproduction

Very simple cra app to test reccomendations
```
import recommend from '@algolia/recommend';
import { useRecommendations } from '@algolia/recommend-react';

import './App.css';

const indexName = 'INDEX_NAME';
const recommendClient = recommend('APP_ ID', 'API_KEY');
const objectId = '99999999';

function App() {
  const { recommendations } = useRecommendations({
      model: 'related-products',
      recommendClient,
      indexName,
      objectIDs: [objectId],
  });

  console.log(recommendations);

  return (
    <div className="App">
      <h1>Testing Algolia Recommend</h1>
    </div>
  );
}

export default App;
```

**Steps**

1. Run above code with correct indices/apps
2. Application enters a full infinite loop constantly logging out recommendations.

## Expected behavior
App should load once with relevant recommendations and not infinitely loop.

## Environment

- OS: macOS
- Browser: Chrome
- Library versions:
    "@algolia/recommend": "^4.10.3",
    "@algolia/recommend-react": "^1.0.0",

## Possible Cause (We have debugged)
Currently the hooks useEffect dependency is reliant on the props object this object contains the `status` prop which is also set within the useEffect thus causing an infinite loop.

shortcuts:
**Summary**

Inlining props in our exposed hooks causes infinite loops: https://github.com/algolia/recommend/issues/49

As of now, users have to provide their own workaround to avoid this kind of issue: https://github.com/algolia/recommend/issues/49#issuecomment-874602027

We are now internally using a new hook called [`useStableValue`](https://github.com/algolia/recommend/blob/fff67727b70d4acf5a3ad99215a3bfbb56d052a5/packages/recommend-react/src/useStableValue.ts) to determine if we should re-render.

**Result**

fixes https://github.com/algolia/recommend/issues/49

Exposed hooks can now be used with inlined props, without causing infinite loops

```ts
  const { recommendations } = useRelatedProducts({
    indexName,
    objectIDs: ['60885-6132-071'],
    recommendClient,
    threshold: 0,
    queryParameters: {
      facetFilters: ['facetName:facetValue'],
    },
    transformItems(items) {
      return items;
    },
  });

```


packages/recommend-react/src/useSafeEffect.ts
```diff
@@ -0,0 +1,55 @@
...
+  const nonPrimitives = Object.entries(props).reduce<TProps>((prev, [k, v]) => {
+    if (!isPrimitive(v)) {
+      return {
+        ...prev,
+        [k]: v,
+      };
+    }
+
+    return prev;
+  }, {} as TProps);
```
Haroenv: these two could be combined in a single reduce if the accumulator was { primitives, nonPrimitives }

shortcuts: Oh right! I mostly focused on make it work, hence the draft, but I'll update it thanks

packages/recommend-react/src/useSafeEffect.ts
```diff
@@ -0,0 +1,55 @@
...
+
+  return { primitives, nonPrimitives };
```
Haroenv:
```suggestion
  return Object.entries(props).reduce<{
    primitives: TProps;
    nonPrimitives: TProps;
  }>(
    (prev, [k, v]) => {
      if (isPrimitive(v)) {
        prev.primitives[k] = v;
      } else {
        prev.nonPrimitives[k] = v;
      }

      return prev;
    },
    { primitives: {} as TProps, nonPrimitives: {} as TProps }
  );
```

mutated here because local mutation in a reduce is safe

packages/recommend-react/src/useStableValue.ts
```diff
@@ -0,0 +1,15 @@
...
+import { useEffect, useState } from 'react';
+
+export function useStableValue<TValue>(value: TValue) {
+  const [stableValue, setStableValue] = useState<TValue>(() => value);
+
+  useEffect(() => {
+    if (dequal(stableValue, value) === false) {
+      setStableValue(value);
+    }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
Haroenv, line 13: this is required because stableValue is not considered "safe" by react right? I wonder if a comment is needed?

Haroenv (approved):
smart solution with not too much overhead!

packages/recommend-react/package.json
```diff
@@ -30,7 +30,8 @@
   },
   "dependencies": {
     "@algolia/recommend-core": "1.0.0",
-    "@algolia/recommend-vdom": "1.0.0"
+    "@algolia/recommend-vdom": "1.0.0",
+    "dequal": "2.0.2"
```
francoischalifour: We don't need to have a hard dependency here, it increases the likelihood of users getting duplicated versions of `dequal` in their projects.

```suggestion
    "dequal": "^2.0.0"
```

packages/recommend-react/src/__tests__/useFrequentlyBoughtTogether.test.ts
```diff
@@ -0,0 +1,50 @@
+import { renderHook } from '@testing-library/react-hooks';
+
+import { createMultiSearchResponse } from '../../../../test/utils/createApiResponse';
+import {
+  createRecommendClient,
+  hit,
+} from '../../../../test/utils/createRecommendClient';
+import { useFrequentlyBoughtTogether } from '../useFrequentlyBoughtTogether';
+
+function createRecommendationsClient() {
```
francoischalifour: Name suggestion: `createRecommendClientMock`

packages/recommend-react/src/__tests__/useFrequentlyBoughtTogether.test.ts
```diff
@@ -0,0 +1,50 @@
...
+          facetFilters: ['test'],
+        },
+        transformItems: (items) => {
+          cb();
+          return items;
+        },
+      });
+
+      expect(recommendations[0]).toEqual(hit);
+      expect(cb).toHaveBeenCalledTimes(1);
```
francoischalifour: That's not related to this test so we can remove this assertion.

packages/recommend-react/src/__tests__/useFrequentlyBoughtTogether.test.ts
```diff
@@ -0,0 +1,50 @@
...
+        queryParameters: {
+          facetFilters: ['test'],
+        },
+        transformItems: (items) => {
+          cb();
+          return items;
+        },
+      });
+
+      expect(recommendations[0]).toEqual(hit);
```
francoischalifour: We can assert the whole `recommendations` with `[hit]`.

packages/recommend-react/src/__tests__/useStableValue.test.tsx
```diff
@@ -0,0 +1,55 @@
...
+        props={{
+          query: 'first query',
+          fn: (value) => value,
+          IDs: ['1', '2', '3'],
+        }}
+      />
+    );
+
+    expect(firstRender.getByText('first query')).toBeInTheDocument();
+    expect(firstRender.getByText(`${42}`)).toBeInTheDocument();
```
francoischalifour:
```suggestion
    expect(firstRender.getByText('42')).toBeInTheDocument();
```

packages/recommend-react/src/__tests__/useStableValue.test.tsx
```diff
@@ -0,0 +1,55 @@
...
+        props={{
+          query: 'second query',
+          fn: (value) => value + 1,
+          IDs: ['1', '2', '3', '4'],
+        }}
+      />
+    );
+
+    expect(firstRender.getByText('second query')).toBeInTheDocument();
+    expect(firstRender.getByText(`${43}`)).toBeInTheDocument();
```
francoischalifour:
```suggestion
    expect(firstRender.getByText('43')).toBeInTheDocument();
```

packages/recommend-react/src/useFrequentlyBoughtTogether.ts
```diff
@@ -6,25 +6,56 @@ import {
...
 import { useStatus } from './useStatus';
 
-export function useFrequentlyBoughtTogether<TObject>(
-  props: GetFrequentlyBoughtTogetherProps<TObject>
-) {
+export function useFrequentlyBoughtTogether<TObject>({
+  objectIDs: userObjectIDs,
+  queryParameters: userQueryParameters,
+  transformItems: userTransformItems,
+
```
francoischalifour: I don't think we need these empty lines?

packages/recommend-react/src/__tests__/useFrequentlyBoughtTogether.test.ts
shortcuts: alternative `createMockedRecommendClient`: https://github.com/algolia/recommend/pull/52/commits/aaf505b1703c2bf816a88dd8d45640256011be2a

packages/recommend-react/src/useStableValue.ts
shortcuts: Indeed, I've added a comment https://github.com/algolia/recommend/pull/52/commits/2c3479e1449111e36298c278b7be35771ceebb96

packages/recommend-react/src/useFrequentlyBoughtTogether.ts
shortcuts: Indeed: https://github.com/algolia/recommend/pull/52/commits/ba9a894f368f107f139e53e60255e53e627ba4a2

packages/recommend-react/src/__tests__/useFrequentlyBoughtTogether.test.ts
shortcuts: removed it https://github.com/algolia/recommend/commit/aaf505b1703c2bf816a88dd8d45640256011be2a

packages/recommend-react/src/__tests__/useFrequentlyBoughtTogether.test.ts
shortcuts: done in https://github.com/algolia/recommend/commit/aaf505b1703c2bf816a88dd8d45640256011be2a

francoischalifour (approved):
Nice! Don't forget to rename the PR before merging (and the commit as well).

PR #54, opened 2021-08-26
chore(dependencies): bump API client version

shortcuts:
**Summary**

Following the bug fix released in the latest version of the Recommend API client ([changelog](https://github.com/algolia/algoliasearch-client-javascript/releases/tag/4.10.5)), we should bump the dependencies to prevent related issues.

### DIFF
--- a/packages/recommend-core/package.json
+++ b/packages/recommend-core/package.json
@@ -28,7 +28,7 @@
     "prepare": "yarn build:esm && yarn build:types"
   },
   "peerDependencies": {
-    "@algolia/recommend": "4.10.5"
+    "@algolia/recommend": "^4.10.5"
   },
   "devDependencies": {
     "@algolia/client-search": "4.9.1"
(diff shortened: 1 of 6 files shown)


packages/recommend-core/package.json
```diff
@@ -28,7 +28,7 @@
     "prepare": "yarn build:esm && yarn build:types"
   },
   "peerDependencies": {
-    "@algolia/recommend": "4.10.2"
+    "@algolia/recommend": "4.10.5"
```
Haroenv:
```suggestion
    "@algolia/recommend": "^4.10.5"
```

should this not be a caret regardless of this PR?

Haroenv (approved):

packages/recommend-core/package.json
shortcuts: Good catch! This might have been missed in the past :D

PR #64, opened 2022-03-10
feat/recommend: Add Trending methods/hooks

PLNech:
- [x] Add `getTrendingItems` and `getTrendingFacets` to _recommend-core_
- [x] Add `getTrendingItems` and `getTrendingFacets` to _recommend-react_
- [ ] Build `TrendingItems` and `TrendingFacets` components on top of these

PR #65, opened 2022-03-10
feat(recommend): Add hooks, methods, react components

SophieManley03:
Take the work from  [this PR](https://github.com/algolia/recommend/pull/64) and add react-components


packages/recommend-vdom/src/FacetsView.tsx
```diff
@@ -0,0 +1,33 @@
...
+    props: FacetsViewProps<TItem, RecommendTranslations, RecommendClassNames>
+  ) {
+    return (
+      <div
+        className={cx('auc-Recommend-container', props.classNames.container)}
+      >
+        <ol className={cx('auc-Recommend-list', props.classNames.list)}>
+          {props.items.map((item) => (
+            <li
+              key={item.facetValue}
```
PLNech, line 23: Should we also use the facetName in the key, so that if we render multiple `TrendingFacets` side by side (e.g. one for `mainColor` one for `secondaryColor`) we don't have two FacetsViews with the same `key`? 

```suggestion
              key={item.facetKey + '_' + item.facetValue}
```

PLNech (approved):
LGTM, nice implementation with the FacetsView! just a small question on the keys of views in `TrendingFacets`, but besides this comment (which I'm unsure of) this is good to go for me :rocket:

packages/recommend-vdom/src/FacetsView.tsx
francoischalifour: React keys are scoped to their list (`map()`) so the facet key shouldn't be needed.

PLNech:
^: Following discussion with @SophieManley03, pushed a little renaming for added clarity :)

packages/recommend-react/src/__tests__/useTrendingItems.test.ts
```diff
@@ -0,0 +1,49 @@
...
+      )
+    ),
+  });
+
+  return {
+    recommendClient,
+  };
+}
+
+describe('useRelatedProducts', () => {
```
francoischalifour:
```suggestion
describe('useTrendingItems', () => {
```

packages/recommend-react/src/__tests__/useTrendingItems.test.ts
```diff
@@ -0,0 +1,49 @@
...
+  test('gets Related Products', () => {
```
francoischalifour:
```suggestion
  test('gets trending items', () => {
```

packages/recommend-react/src/__tests__/useTrendingFacets.test.ts
```diff
@@ -0,0 +1,48 @@
...
+      )
+    ),
+  });
+
+  return {
+    recommendClient,
+  };
+}
+
+describe('useRelatedProducts', () => {
```
francoischalifour:
```suggestion
describe('useTrendingFacets', () => {
```

packages/recommend-react/src/__tests__/useTrendingFacets.test.ts
```diff
@@ -0,0 +1,48 @@
...
+  test('gets Related Products', () => {
```
francoischalifour:
```suggestion
  test('gets trending facets', () => {
```

packages/recommend-core/src/utils/mapToTrendingFacets.ts
```diff
@@ -0,0 +1,31 @@
...
+import { TrendingFacet } from '../types';
+
+import { sortBy } from './sortBy';
+
+type MapToRecommendations<TObject> = {
+  response: MultipleQueriesResponse<TObject>;
+  maxRecommendations?: number;
+};
+
+export function mapToTrendingFacets<TObject>({
```
francoischalifour: Could we avoid this duplicated logic (with `mapToRecommendations`)? It'd be possible by accepting `hits` instead of `response` and using `uniqBy` one call higher.

francoischalifour (approved):

packages/recommend-core/src/utils/mapToTrendingFacets.ts
SophieManley03: I tried what you are saying. And it's working, Thank you for your advice :) (If I understood correctly)

packages/recommend-core/src/getFrequentlyBoughtTogether.ts
```diff
@@ -31,9 +32,14 @@ export function getFrequentlyBoughtTogether<TObject>({
   return recommendClient
     .getFrequentlyBoughtTogether<TObject>(queries)
     .then((response) =>
-      mapToRecommendations({
+      // Multiple identical recommended `objectID`s can be returned b
```
francoischalifour: Can we fix this typo while we're at it? (same applies below)

```suggestion
      // Multiple identical recommended `objectID`s can be returned by
```

packages/recommend-core/src/getRecommendations.ts
```diff
@@ -58,9 +58,14 @@ export function getRecommendations<TObject>({
   return recommendClient
     .getRecommendations<TObject>(queries)
     .then((response) =>
-      mapToRecommendations({
+      // Multiple identical recommended `objectID`s can be returned b
```
francoischalifour:
```suggestion
      // Multiple identical recommended `objectID`s can be returned by
```

packages/recommend-core/src/getRelatedProducts.ts
```diff
@@ -31,9 +32,14 @@ export function getRelatedProducts<TObject>({
   return recommendClient
     .getRelatedProducts<TObject>(queries)
     .then((response) =>
-      mapToRecommendations({
+      // Multiple identical recommended `objectID`s can be returned b
```
francoischalifour:
```suggestion
      // Multiple identical recommended `objectID`s can be returned by
```

francoischalifour (approved):

packages/recommend-core/src/getFrequentlyBoughtTogether.ts
```diff
@@ -31,9 +32,14 @@ export function getFrequentlyBoughtTogether<TObject>({
   return recommendClient
     .getFrequentlyBoughtTogether<TObject>(queries)
     .then((response) =>
-      mapToRecommendations({
+      // Multiple identical recommended `objectID`s can be returned by
+      // the engine, so we need to remove duplicates.
```
Haroenv: can that be solved on the API side in the future?

packages/recommend-core/src/getTrendingFacets.ts
```diff
@@ -0,0 +1,61 @@
...
+    facetName,
+  };
+
+  recommendClient.addAlgoliaAgent('recommend-core', version);
+
+  return recommendClient
+    .getTrendingFacets<TObject>([query])
+    .then((response) =>
+      // Multiple identical recommended `objectID`s can be returned b
+      // the engine, so we need to remove duplicates.
```
Haroenv: there's the same uniqBy comment as in the other places, but uniqBy isn't called. Is that expected?

Haroenv: the comment should in either case be moved two lines lower to be about the hits, right?

packages/recommend-core/src/getTrendingItems.ts
```diff
@@ -0,0 +1,66 @@
...
+    facetValue,
+  };
+
+  recommendClient.addAlgoliaAgent('recommend-core', version);
+
+  return recommendClient
+    .getTrendingItems<TObject>([query])
+    .then((response) =>
+      // Multiple identical recommended `objectID`s can be returned b
+      // the engine, so we need to remove duplicates.
```
Haroenv: comment should be moved two lines lower

Haroenv (commented):
👍 great job!

packages/recommend-core/src/getTrendingFacets.ts
SophieManley03: Yes I agree 👍

Haroenv (approved):

SophieManley03:
Thank you for the reviews 🚀

packages/recommend-core/src/getFrequentlyBoughtTogether.ts
SophieManley03: cc @jojva for visibility

jojva: I'm not really aware of this limitation, and knowing the code I'm actually surprised that would be the case.

jojva: I'll have to dig that a bit

jojva: I confirm we never return duplicated `objectID`s, you can remove this logic.

PR #71, opened 2022-03-21
feat(recommend): Add trending for recommend-js

SophieManley03:
Trending has been added to recommend and recommend-react. We also needed to upgrade recommend-js.


PLNech (approved):
LGTM! :rocket:

marialungu (approved):

PR #77, opened 2022-04-04
Fix(core): remove sorting on recommendations retrieval

marialungu:
Recommendations were being sorted and this affects the behaviour of Recommend rules (order is given from the API).
Removed filtering based on 'objectID' as it is no longer needed.


francoischalifour:
This was necessary when providing multiple `objectIDs` like recommendations in a cart. Doesn't this break this behavior?

marialungu:
HI @francoischalifour,

@jojva confirmed that they don't return duplicated `objectID`s in a previous [PR](https://github.com/algolia/recommend/pull/65#discussion_r830972535), so I think we can safely remove this logic.

francoischalifour:
Right, but do they sort the merged hits now? https://github.com/algolia/recommend/pull/77/files#diff-7f612e4ac1bfa038efb03b9dc0e7bf5c9ef71b834c37d522c36b24c49f049c08L12-L19

jojva:
Hi @francoischalifour,

> Right, but do they sort the merged hits now?

That's the part I don't understand. When we say "multiple indices", does this mean a recommendation request with several sub-queries, each targeting an index? And if so, I wonder why the library needs to re-sort those, discarding the origin of the index?

Results do not necessarily return a recommendation score, as we now have recommend rules, that are able to pin items. And secondly, recommendation scores are not necessarily the only basis for the ranking. It can also be affected by boost and bury (new recommend rules options), as well as the rest of the ranking formula.

I don't think we can or should mix hits coming from several indices. I'd like to understand the idea behind this and how we can solve the issue.

francoischalifour:
@jojva I'm out of the loop with new Recommend features like rules and pins. How about we sync together this week? DMing you.

marialungu:
Closing as this issue needs to be discussed further given its impact on the recommend flows.

PR #78, opened 2022-04-14
Fix(core): conflicting rules (RECO-523)

marialungu:
Added new function for merging and sorting related-products and freq-bought-together taking into account applied rules by sorting based on average index of object occurrences

### DIFF
--- a/packages/recommend-core/src/getRecommendations.ts
+++ b/packages/recommend-core/src/getRecommendations.ts
@@ -1,7 +1,7 @@
 import type { RecommendClient, RecommendationsQuery } from '@algolia/recommend';
 
 import { ProductRecord, RecordWithObjectID } from './types';
-import { mapToRecommendations, uniqBy } from './utils';
+import { mapToRecommendations } from './utils';
 import { version } from './version';
 
 export type RecommendationsProps<TObject> = {
--- a/packages/recommend-core/src/getRecommendations.ts
+++ b/packages/recommend-core/src/getRecommendations.ts
@@ -62,10 +62,8 @@ export function getRecommendations<TObject>({
         maxRecommendations,
         // Multiple identical recommended `objectID`s can be returned b
         // the engine, so we need to remove duplicates.
-        hits: uniqBy<ProductRecord<TObject>>(
-          'objectID',
-          response.results.map((result) => result.hits).flat()
-        ),
+        hits: response.results.map((result) => result.hits),
+        nrOfObjs: objectIDs.length,
       })
     )
     .then((hits) => ({ recommendations: transformItems(hits) }));
--- a/packages/recommend-core/src/getRecommendations.ts
+++ b/packages/recommend-core/src/getRecommendations.ts
@@ -60,8 +60,6 @@ export function getRecommendations<TObject>({
     .then((response) =>
       mapToRecommendations<ProductRecord<TObject>>({
         maxRecommendations,
-        // Multiple identical recommended `objectID`s can be returned b
-        // the engine, so we need to remove duplicates.
         hits: response.results.map((result) => result.hits),
         nrOfObjs: objectIDs.length,
       })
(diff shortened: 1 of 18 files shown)


packages/recommend-core/src/getRecommendations.ts
```diff
@@ -62,10 +62,8 @@ export function getRecommendations<TObject>({
         maxRecommendations,
         // Multiple identical recommended `objectID`s can be returned b
         // the engine, so we need to remove duplicates.
-        hits: uniqBy<ProductRecord<TObject>>(
-          'objectID',
-          response.results.map((result) => result.hits).flat()
-        ),
+        hits: response.results.map((result) => result.hits),
```
SophieManley03, line 63: You can remove the comments as you did above. (Same comments for `getRelatedProducts`)

packages/recommend-core/src/getTrendingFacets.ts
```diff
@@ -1,7 +1,7 @@
 import { RecommendClient, TrendingFacetsQuery } from '@algolia/recommend';
 
 import { TrendingFacet } from './types';
-import { mapToRecommendations } from './utils';
+import { mapWithSortToRecommendations } from './utils';
```
SophieManley03: Here I'm thinking that both methods `mapToRecommendations` and `mapWithSortToRecommendations` are sorting the results. Maybe we could provide a more explicit naming for the latest. For example `mapByScoreToRecommendation`? WDYT?
(This is nitpicking, feel free to keep as is)

packages/recommend-core/src/utils/computeAverageIndexes.ts
```diff
@@ -0,0 +1,21 @@
+import { AverageIndexes } from '../types/AverageIndexes';
+import { IndexTracker } from '../types/IndexTracker';
+
+export const getAverageIndexes = (
+  indexTracker: IndexTracker,
+  nrOfObjs: number
```
SophieManley03, line 8: Maybe it's worth documented the props? When I first read this file I wasn't sure what `nrOfObjs` was.

packages/recommend-core/src/utils/mapToRecommendations.ts
```diff
@@ -1,22 +1,55 @@
...
 }: MapToRecommendations<TObject>) {
-  // Since recommendations from multiple indices are returned, we
-  // need to sort them descending based on their score.
-  return sortBy<TObject & { _score?: number }>((a, b) => {
-    const scoreA = a._score || 0;
-    const scoreB = b._score || 0;
+  const indexTracker: IndexTracker = {};
+
+  // eslint-disable-next-line array-callback-return
+  hits.map((arr) => {
```
SophieManley03: If you don't return anything you can use a forEach :). Like that you might remove `eslint-disable-next-line array-callback-return`

marialungu: ah good point, thank you

SophieManley03 (approved):
Good job 👍

packages/recommend-core/src/utils/computeAverageIndexes.ts
```diff
@@ -3,6 +3,8 @@ import { IndexTracker } from '../types/IndexTracker';
 
 export const getAverageIndexes = (
```
francoischalifour, line 4: Nit: we use the term "indices" instead of "indexes" everywhere in our product communication/documentation.

packages/recommend-core/src/utils/mapToRecommendations.ts
```diff
@@ -13,20 +13,20 @@ type MapToRecommendations<TObject> = {
 export function mapToRecommendations<TObject>({
   hits,
   maxRecommendations,
+  // total number of products for which we are retrieving recommendations
+  // e.g. objectsIDs.length
```
francoischalifour, line 17: We can use TSDoc in the `MapToRecommendations` type to avoid having multiple definitions of the same param.

PR #79, opened 2022-04-20
fix(core): cleanup prev fix for rules

marialungu:
Renaming 'indexes' -> 'indices' for consistency
Added TSDoc for `mapToRecommendations` fction


packages/recommend-core/src/utils/mapToRecommendations.ts
```diff
@@ -13,10 +13,16 @@ type MapToRecommendations<TObject> = {
 export function mapToRecommendations<TObject>({
   hits,
   maxRecommendations,
-  // total number of products for which we are retrieving recommendations
-  // e.g. objectsIDs.length
   nrOfObjs,
 }: MapToRecommendations<TObject>) {
+  /**
```
SophieManley03: This can be placed above the `function` l.12

SophieManley03 (approved):

packages/recommend-core/src/utils/mapToRecommendations.ts
```diff
@@ -1,8 +1,16 @@
...
-import { getAverageIndexes } from './computeAverageIndexes';
+import { getAverageIndices } from './computeAverageIndices';
+
+/**
+ * Returns the reordered list of recommendations based on average indices.
+ *
+ * @param hits - recommendations
+ * @param maxRecommendations - max number of recommendations
+ * @param nrOfObjs - total number of products for which we are retrieving recommendations (objectsIDs.length)
+ */
```
francoischalifour, line 13: I meant like this: https://github.com/algolia/recommend/blob/7d9bb18046f715e8bcf61bf84ae0dbf298ba095a/packages/recommend-core/src/getTrendingItems.ts#L7-L20

PR #91, opened 2022-06-27
feat(rendering): add option to remove header when no recos are returned

marialungu:
### **What**
Add the option for the user to not display the headerComponent when there are no recommendations returned 

### **How**
Add a new parameter `removeHeaderComponent` -> if set to true, then the headerComponent will not be displayed when there are no recommendations returned

### DIFF
--- a/packages/recommend-vdom/src/DefaultHeader.tsx
+++ b/packages/recommend-vdom/src/DefaultHeader.tsx
@@ -5,6 +5,13 @@ import { cx } from './utils';
 
 export function createDefaultHeaderComponent({ createElement }: Renderer) {
   return function DefaultHeader<TObject>(props: ComponentProps<TObject>) {
+    if (
+      (!props.recommendations || props.recommendations.length < 1) &&
+      props.removeHeaderComponent
+    ) {
+      return null;
+    }
+
     if (!props.translations.title) {
       return null;
     }
--- a/packages/recommend-vdom/src/DefaultHeader.tsx
+++ b/packages/recommend-vdom/src/DefaultHeader.tsx
@@ -5,10 +5,7 @@ import { cx } from './utils';
 
 export function createDefaultHeaderComponent({ createElement }: Renderer) {
   return function DefaultHeader<TObject>(props: ComponentProps<TObject>) {
-    if (
-      (!props.recommendations || props.recommendations.length < 1) &&
-      props.removeHeaderComponent
-    ) {
+    if (!props.recommendations || props.recommendations.length < 1) {
       return null;
     }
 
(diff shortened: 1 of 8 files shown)


francoischalifour:
Are there any scenarios where showing the header without recommendations is desirable? If not, why introduce a new option?

I don't think we should introduce new combinatory options.

marialungu:
Hi @francoischalifour,

here's the slack [conversation](https://algolia.slack.com/archives/C01V22PTKB9/p1655975367469889),

The idea is to give the option to remove the comp header when you don't have a fallback comp and no recommendations are being returned. You can either:

- leave it as is (no recos + no feedback => empty comp with header) (with our default components)
- `removeHeaderComponent = {true}` => you remove the header Comp when there are no recommendations  (with our default components)
- use custom header comp and you handle this in your own app

packages/recommend-vdom/src/DefaultHeader.tsx
```diff
@@ -5,6 +5,13 @@ import { cx } from './utils';
 
 export function createDefaultHeaderComponent({ createElement }: Renderer) {
   return function DefaultHeader<TObject>(props: ComponentProps<TObject>) {
+    if (
+      (!props.recommendations || props.recommendations.length < 1) &&
+      props.removeHeaderComponent
```
SophieManley03: Maybe the name is not clear enough as this will only remove the header when there is no results and if the user is not passing a `headerComponent`. WDYT?

francoischalifour:
@marialungu I understand what this PR does. What I'm suggesting is that this probably shouldn't be a new option, but the default behavior. I can't think of a reason why you'd want the header displayed when there's nothing to show in the section.

And in a more general API design perspective, we should limit as much as we can boolean options.

marialungu:
@francoischalifour,

ah, sorry for the misunderstanding, i rushed it a bit 🙈
I agree with your suggestion, thank you for your help

SophieManley03 (approved):

packages/recommend-vdom/src/DefaultHeader.tsx
```diff
@@ -5,6 +5,10 @@ import { cx } from './utils';
 
 export function createDefaultHeaderComponent({ createElement }: Renderer) {
   return function DefaultHeader<TObject>(props: ComponentProps<TObject>) {
+    if (!props.recommendations || props.recommendations.length < 1) {
```
francoischalifour, line 8: `props.recommendations` is always an array so why this first guard? We can also merge it with the following condition?

```suggestion
    if (props.recommendations.length === 0 || !props.translations.title) {
```

PR #98, opened 2022-07-27
feat(recommend) HTML template 

SophieManley03:
Following the work done in autocomplete we want to enable HTML templating into recommend-js.


packages/recommend-vdom/src/FacetsView.tsx
```diff
@@ -23,11 +23,11 @@ export function createFacetsView({ createElement, Fragment }: Renderer) {
               key={item.facetValue}
               className={cx('auc-Recommend-item', props.classNames.item)}
             >
-              <props.itemComponent
-                createElement={createElement}
-                Fragment={Fragment}
-                item={item}
-              />
+              {props.itemComponent({
```
SophieManley03: Using it as a function allow to return VNode[]. It is not possible with the <props.itemComponent /> syntax. 

(Error triggered: `  Its return type 'VNode<any> | VNode<any>[]' is not a valid JSX element.`)

francoischalifour: I think this can cause problems because it means `createElement()` won't get called, and the underlying framework won't be able to interact with this function. For instance, you won't be able to use React Context in there. That could be problematic if users use a `<Link>` from Next.js in their item component.

Not 100% sure but we should try that first.

packages/recommend-js/src/types/html.tsx
```diff
@@ -0,0 +1,5 @@
+import { VNode } from '@algolia/recommend-vdom';
```
aymeric-giraudet: You can actually just use `import { html } from 'htm/preact` which does exactly this and [simply returns a `VNode`](https://github.com/developit/htm/blob/d62dcfdc721e47bc1923a2cb7a01ebd594ab0c25/src/integrations/preact/index.d.ts#L4)

aymeric-giraudet: Which means this solves the issue with `VNode | VNode[]` :)

SophieManley03: I noticed that in autocomplete, we use this VNode | VNode[]. That is why I wanted to use VNode[], to have the same type accross our lib. In the meantime you are right and it does solve our issue. 🤔

aymeric-giraudet: Indeed in autocomplete templates are considered just as functions that return vdom elements.
In recommend we can pass components that will be instantiated via `createElement` or `h` depending on whether we used React or Preact, which I think is even better as it allows better debugging via devtools, proper usage of hooks, and allows users to fine-tune rerendering 😄 
That way we also have to change less pre-existing code

packages/recommend-js/src/types/HTMLTemplate.tsx
```diff
@@ -0,0 +1,5 @@
+import { html } from 'htm/preact';
+
+export type Template = {
```
sarahdayan:
```suggestion
export type HTMLTemplate = {
```

examples/templating/htmlTemplating/index.html
```diff
@@ -0,0 +1,40 @@
+<html lang="en">
```
sarahdayan: Should we rename the `htmlTemplating` directory into `html`? The term "templating" is redundant with the parent folder.

packages/recommend-js/src/frequentlyBoughtTogether.tsx
```diff
@@ -71,15 +77,45 @@ function FrequentlyBoughtTogether<TObject>(
...
   if (!container) {
-    return children;
+    return vnode;
   }
 
-  render(children, getHTMLElement(container, environment));
+  render(vnode, getHTMLElement(container, environment));
 
-  return undefined;
+  return null;
```
francoischalifour, line 108: Why did we change this return type?

packages/recommend-js/src/relatedProducts.tsx
```diff
@@ -61,15 +68,45 @@ function RelatedProducts<TObject>(props: RelatedProductsProps<TObject>) {
...
+      headerComponent={
+        headerComponent
+          ? (headerProps) => headerComponent({ ...headerProps, html })
+          : undefined
+      }
+      fallbackComponent={
+        fallbackComponent
+          ? (fallbackProps) => fallbackComponent({ ...fallbackProps, html })
+          : undefined
+      }
```
francoischalifour: Could we create a function that injects `html` in the component props?

```tsx
import { html } from 'htm/preact';

function injectHtmlProp(component) {
  return (props) => component({ ...props, html });
}
```

```jsx
<TrendingItems
  {...props}
  view={view && injectHtmlProp(view)}
  itemComponent={itemComponent && injectHtmlProp(itemComponent)}
  headerComponent={headerComponent && injectHtmlProp(headerComponent)}
  fallbackComponent={fallbackComponent && injectHtmlProp(fallbackComponent)}
/>
```

Maybe something like this?

Then it's easier to read imo, and we can use it in every component.

francoischalifour (commented):
Great additions!

packages/recommend-js/src/frequentlyBoughtTogether.tsx
SophieManley03: It seems more accurate to return null when we have nothing in that case. Plus, we show examples where in the fallbackComponent we use one of our component in recommend but the types are incompatible as one is null and the other undefined.

packages/recommend-js/src/relatedProducts.tsx
SophieManley03: Indeed it is more readable!

aymeric-giraudet: I'm sorry, this made me realize that I forgot in my PR to reference the components as JSX elements here too.
It rather looks to me like it's a HOC, we could have : 
```tsx
function withHtml<TProps>(
  Component: preact.FunctionComponent<TProps>
): (props: TProps & Template) => JSX.Element {
  const ComponentWithHtml = (props: TProps) => (
    <Component {...props} html={html} />
  );

  ComponentWithHtml.displayName = `withHtml(${Component.displayName})`;

  return ComponentWithHtml;
}
```
That way we have the same behavior as before this PR (passed components are called with Preact's `h` function), [are less prone to bugs](https://kentcdodds.com/blog/dont-call-a-react-function-component) and allow for clearer debugging, when we pass a ProductItem component directly to itemComponent, here are Preact devtools : 
image
Top is current implementation, it shows ProductItem as if it was a custom hook which is weird, and doesn't show the `html` prop.
At the bottom is with the HOC, it properly shows the name of the component which I think users will appreciate

What do you both think about it ?

francoischalifour: @aymeric-giraudet That's much better indeed!

aymeric-giraudet:
Looks good ! I think it would be nice to add some tests now and it'll be good to go

packages/recommend-js/src/utils/withHtml.tsx
```diff
@@ -0,0 +1,17 @@
...
+import { HTMLTemplate } from '../types';
+
+export function withHtml<TProps>(
+  Component: preact.FunctionComponent<TProps>
+): (props: TProps & HTMLTemplate) => JSX.Element {
+  const ComponentWithHtml = (props: TProps) => (
+    <Component {...props} html={html} />
+  );
+
+  ComponentWithHtml.displayName = `withHtml(${Component.displayName})`;
```
francoischalifour: We can save some bytes on the production bundle:

```suggestion
  if (__DEV__) {
    ComponentWithHtml.displayName = `withHtml(${Component.displayName})`;
  }
```

packages/recommend-js/src/utils/withHtml.tsx
```diff
@@ -11,7 +11,7 @@ export function withHtml<TProps>(
     <Component {...props} html={html} />
   );
 
-  if (__DEV__) {
+  if (process.env.__DEV__) {
```
francoischalifour: This won't get picked up by Babel and Rollup. I thought we had the tooling to support `__DEV__` in this repo already.

You need these additions:

- https://github.com/algolia/predict/blob/4537902a610cbc972a5e1a39a885997a0bc38de0/packages/predict-core/global.d.ts
- https://github.com/algolia/predict/blob/4537902a610cbc972a5e1a39a885997a0bc38de0/babel.config.js#L36-L45

But let's do that in another PR then.

sarahdayan: @SophieManley03 I'd recommend doing this in a separate PR now, merge it to `next`, and rebase this branch.

SophieManley03: Sure I can do it now. 
The purpose of it, is to make the production bundle smaller?

francoischalifour: Yes, this way users don't pay the bundle price on their production apps.

SophieManley03: Thank you for the explanation, I did the update ✅ .

examples/templating/html/index.ts
```diff
@@ -0,0 +1,75 @@
...
+  itemComponent: productItem,
+  view({ items, itemComponent, html, createElement, Fragment }) {
+    return horizontalSlider({
+      items,
+      itemComponent({ item }) {
+        return itemComponent({ item, html, createElement, Fragment });
+      },
+    });
+  },
+});
```
Haroenv:
```suggestion
  view({items, ...renderer}) {
    return horizontalSlider({
      items,
      itemComponent({ item }) {
        return itemComponent({ item, ...renderer });
      },
    });
  },
});
```

I wonder if a pattern like this could make sense so that html, createElement and Fragment don't need to be repeated

packages/recommend-js/src/__tests__/frequentlyBoughtTogether.test.tsx
```diff
@@ -43,51 +107,12 @@ describe('frequentlyBoughtTogether', () => {
...
-                    <div>
-                      3
-                    </div>
-                  </li>
-                </ol>
-              </div>
-            </section>
-          </div>
-        `);
+        expect(container).toMatchInlineSnapshot(headerAndItemsComponent);
```
Haroenv: Normally an inline snapshot is completely inline, I haven't seen it extracted to a variable before, why is that done?

packages/recommend-js/src/utils/withHtml.tsx
```diff
@@ -0,0 +1,19 @@
+/** @jsx h */
+import { html } from 'htm/preact';
+import { h } from 'preact';
+
+import { HTMLTemplate } from '../types';
+
+export function withHtml<TProps>(
```
Haroenv, line 7: as this imports html statically, it could also be `useHtml` instead of a HOC.

It wouldn't really be a hook, just a function that receives something, so `getHtml` or `html` exported from here could work too

Haroenv (commented):
👍

packages/recommend-js/src/__tests__/frequentlyBoughtTogether.test.tsx
SophieManley03: I wanted to avoid repeating the same section of code

packages/recommend-js/src/utils/withHtml.tsx
SophieManley03: Thanks for providing the alternative. 👌 
For now I would keep this solution as it is working as is

Haroenv (approved):
🙆‍♂️ looks correct!

SophieManley03:
Thank you all for your help 🙇‍♀️

PR #103, opened 2022-08-31
chore: Expose __DEV__

SophieManley03:


package.json
```diff
@@ -43,6 +43,7 @@
     "@typescript-eslint/eslint-plugin": "2.19.0",
     "@typescript-eslint/parser": "2.19.0",
     "babel-eslint": "10.1.0",
+    "babel-plugin-inline-replace-variables": "^1.3.1",
```
francoischalifour: Dev dependencies are pinned (and need to re-run `yarn install`).

```suggestion
    "babel-plugin-inline-replace-variables": "1.3.1",
```

sarahdayan: @SophieManley03 If you applied from this suggestion you need to re-run `yarn` locally to re-generate `yarn.lock`. This is why your build is currently failing.

sarahdayan (changes requested):
There shouldn't be two lockfiles. Make sure to always use `yarn` to avoid generating `package-lock.json`.

francoischalifour (approved):

sarahdayan (approved):

PR #109, opened 2022-09-30
feat(react): accept initial results (RECO-722)

marialungu:
Accept `initialResult` in order to enable the further implementation of SSR in Recommend (as discussed in [RFC](https://algolia.atlassian.net/wiki/spaces/RECO/pages/4276355148/RFC+-+react-recommend+-+Expose+initialState))


packages/recommend-react/src/useAlgoliaAgent.ts
```diff
@@ -5,11 +5,15 @@ import { version } from './version';
...
 >;
 
 export function useAlgoliaAgent(props: UseAlgoliaAgentProps) {
   useEffect(() => {
-    props.recommendClient.addAlgoliaAgent('recommend-react', version);
-  }, [props.recommendClient]);
+    if (props.initialState) {
+      props.recommendClient.addAlgoliaAgent('recommend-server', version);
+    } else {
+      props.recommendClient.addAlgoliaAgent('recommend-react', version);
```
francoischalifour: We need to add the `recommend-react` user agent either way.

packages/recommend-react/src/useAlgoliaAgent.ts
```diff
@@ -5,11 +5,15 @@ import { version } from './version';
...
-  'recommendClient'
+  'recommendClient' | 'initialState'
 >;
 
 export function useAlgoliaAgent(props: UseAlgoliaAgentProps) {
   useEffect(() => {
-    props.recommendClient.addAlgoliaAgent('recommend-react', version);
-  }, [props.recommendClient]);
+    if (props.initialState) {
+      props.recommendClient.addAlgoliaAgent('recommend-server', version);
```
francoischalifour: I'm not sure that's a correct heuristic. You could provide `initialState` and not be on the server.

packages/recommend-core/src/getTrendingFacets.ts
```diff
@@ -17,6 +18,12 @@ export type TrendingFacetsProps<TObject> = {
   transformItems?: (
     items: Array<TrendingFacet<TObject>>
   ) => Array<TrendingFacet<TObject>>;
+  /**
+   * A prop that allows the user to provide initial results from outside
+   *
+   * It's useful for building SSR on top of Recommend
+   */
+  initialState?: InitialResults<TObject>;
```
francoischalifour: This prop shouldn't be part of the static get methods, but only in our Hooks.

packages/recommend-react/src/useFrequentlyBoughtTogether.ts
```diff
@@ -17,10 +17,12 @@ export function useFrequentlyBoughtTogether<TObject>({
   recommendClient,
   threshold,
   transformItems: userTransformItems,
+  initialState,
 }: GetFrequentlyBoughtTogetherProps<TObject>) {
-  const [result, setResult] = useState<GetRecommendationsResult<TObject>>({
-    recommendations: [],
-  });
+  const initialResults = initialState ?? { recommendations: [] };
```
francoischalifour: We should align the "results" and "state" naming. We seem to use `initialState` and `initialResults` interchangeably.

francoischalifour (changes requested):
The initial `useEffect()` call should be skipped on first render. Otherwise you SSR but once the library loads on the browser you still trigger a network call.

We need to add tests:

- no calls on the browser with `initialState`
- correct user agents

packages/recommend-core/src/getTrendingFacets.ts
marialungu: ✅

packages/recommend-react/src/useFrequentlyBoughtTogether.ts
```diff
@@ -28,23 +31,26 @@ export function useFrequentlyBoughtTogether<TObject>({
...
-      objectIDs,
-      queryParameters,
-      recommendClient,
-      threshold,
-      transformItems,
-    }).then((response) => {
-      setResult(response);
-      setStatus('idle');
-    });
+    if (!initialState) {
```
francoischalifour: You need to wrap the whole effect in this condition, otherwise you get on a stuck "loading" state.

Also, this effect should be skipped only on first execution. Otherwise you won't ever refetch recommendations when one of the props change.

(I would advise to write all test scenarios, because SSR support complexifies the flow.)

marialungu: ah great thank you, was just about to ask this 😄

test/utils/createRecommendClient.ts
```diff
@@ -33,6 +33,10 @@ export const hit = {
   ],
 };
 
+export const initialState = {
+  recommendations: [hit],
+};
```
francoischalifour: This could be in a `test/fixtures` file, but this doesn't belong to the API response.

packages/recommend-react/src/__tests__/useFrequentlyBoughtTogether.test.ts
```diff
@@ -1,9 +1,15 @@
+import { waitFor } from '@testing-library/dom';
 import { renderHook } from '@testing-library/react-hooks';
+import React from 'react';
 
-import { createMultiSearchResponse } from '../../../../test/utils/createApiResponse';
+import {
+  createMultiSearchResponse,
+  forceDelay,
```
francoischalifour: This shouldn't be part of the API response either.

packages/recommend-react/src/__tests__/useFrequentlyBoughtTogether.test.ts
```diff
@@ -24,11 +30,11 @@ function createMockedRecommendClient() {
 }
 
 describe('useFrequentlyBoughtTogether', () => {
-  test('gets Frequently Bought Together products', () => {
+  test('gets Frequently Bought Together products', async () => {
```
francoischalifour: Let's try to use the Recommend vocabulary consistently.

```suggestion
  test('returns FBT recommendations', async () => {
```

packages/recommend-react/src/__tests__/useFrequentlyBoughtTogether.test.ts
```diff
@@ -37,9 +43,98 @@ describe('useFrequentlyBoughtTogether', () => {
...
-      });
+      })
+    );
 
-      expect(recommendations).toEqual([hit]);
+    await waitFor(() => {
+      expect(result.current.recommendations).toEqual([hit]);
     });
   });
+  test('gets fbt products from initialState', async () => {
```
francoischalifour:
```suggestion

  test('returns FBT recommendations from initialState', async () => {
```

packages/recommend-react/src/__tests__/useFrequentlyBoughtTogether.test.ts
```diff
@@ -37,9 +43,98 @@ describe('useFrequentlyBoughtTogether', () => {
...
+    );
+
+    await waitFor(() => {
+      expect(result.current.recommendations).toEqual(
+        initialState.recommendations
+      );
+    });
+  });
+
+  test('no network calls with initialState', () => {
```
francoischalifour:
```suggestion
  test("doesn't initially fetch the recommendations with initialState", () => {
```

packages/recommend-react/src/__tests__/useFrequentlyBoughtTogether.test.ts
```diff
@@ -37,9 +43,98 @@ describe('useFrequentlyBoughtTogether', () => {
...
+        wrapper: React.StrictMode,
+      }
+    );
+
+    expect(recommendClient.getFrequentlyBoughtTogether).toHaveBeenCalledTimes(
+      0
+    );
+  });
+
+  test('trigger network call when props change', async () => {
```
francoischalifour:
```suggestion
  test('fetches recommendations when props change with initialState', async () => {
```

francoischalifour: Ideally we should have this test with all the props in the `useEffect()` deps array.

packages/recommend-react/src/__tests__/useFrequentlyBoughtTogether.test.ts
```diff
@@ -37,9 +43,98 @@ describe('useFrequentlyBoughtTogether', () => {
...
+      {
+        wrapper: React.StrictMode,
+        initialProps: { indexName: 'test' },
+      }
+    );
+    expect(recommendClient.getFrequentlyBoughtTogether).toHaveBeenCalledTimes(
+      0
+    );
+
+    await forceDelay(1000);
```
francoischalifour: Why do we need to have a delay? Leaving context either as a code comment or in this PR would help reviewers understand this.

packages/recommend-react/src/__tests__/useRecommendations.test.ts
```diff
@@ -24,11 +30,11 @@ function createMockedRecommendClient() {
 }
 
 describe('useRecommendations', () => {
-  test('gets recommendations', () => {
+  test('gets recommendations', async () => {
```
francoischalifour: We can apply the same vocabulary in all the other tests as well.

packages/recommend-react/src/useFrequentlyBoughtTogether.ts
```diff
@@ -17,32 +18,43 @@ export function useFrequentlyBoughtTogether<TObject>({
...
+        transformItems,
+      }).then((response) => {
+        setResult(response);
+        setStatus('idle');
+      });
+    } else {
+      renderRef.current = true;
+    }
   }, [
+    initialState,
```
francoischalifour: We should stabilize the `initialState` above with `useStableValue()`, otherwise if users inline their `initialState`, it'll trigger an infinite loop.

packages/recommend-react/src/useFrequentlyBoughtTogether.ts
```diff
@@ -17,32 +18,43 @@ export function useFrequentlyBoughtTogether<TObject>({
...
-  });
+  initialState,
+}: GetFrequentlyBoughtTogetherProps<TObject> & {
+  initialState?: InitialResults<TObject>;
+}) {
+  const initialResults = initialState ?? { recommendations: [] };
+  const [result, setResult] = useState<GetRecommendationsResult<TObject>>(
+    initialResults
+  );
+  const renderRef = useRef(false);
```
francoischalifour: We could make this ref more explicit with something like this:

```suggestion
  const isFirstRenderRef = useRef(true);
```

packages/recommend-react/src/useFrequentlyBoughtTogether.ts
francoischalifour: @marialungu Any updates on this?

packages/recommend-react/src/useAlgoliaAgent.ts
francoischalifour: @Haroenv @sarahdayan Do you have opinions about this?

Haroenv: the server agent indeed should be added on the place that _sends_ the request, not the place that _receives_ the initialState

packages/recommend-react/src/__tests__/useRecommendations.test.ts
marialungu: ✅

packages/recommend-react/src/__tests__/useFrequentlyBoughtTogether.test.ts
marialungu: ✅

packages/recommend-react/src/__tests__/useFrequentlyBoughtTogether.test.ts
marialungu: ✅

packages/recommend-react/src/useFrequentlyBoughtTogether.ts
marialungu: yeah, i'll include it with the next refactors

packages/recommend-react/src/useAlgoliaAgent.ts
sarahdayan: Yes, I think this is too early to add this agent. The PR isn't about SSR, but about initial state. Even if the initial state API is used by SSR, we're mixing concerns here that are unrelated.

packages/recommend-react/src/__tests__/useFrequentlyBoughtTogether.test.ts
marialungu: ✅

examples/server-demo/src/server.js
```diff
@@ -0,0 +1,68 @@
...
+  const initResults = {
+    recommendations: [
+      {
+        name: 'Sneakers Hogan Rebel pink',
+        objectID: 'MX1234',
+        price: {
+          value: 278,
+          currency: 'EUR',
+        },
+        // eslint-disable-next-line @typescript-eslint/camelcase
```
francoischalifour: You could use the instruction at the file level only once, at the top:

```js
/* eslint-disable @typescript-eslint/camelcase */
```

examples/server-demo/src/server.js
```diff
@@ -0,0 +1,68 @@
...
+          'https://res.cloudinary.com/hilnmyskv/image/upload/v1638372051/flagship_sunrise/M0E20000000DWIV_0.jpg',
+        ],
+      },
+    ],
+  };
+
+  res.send(
+    html({
+      body: app,
+      initialState: initResults,
```
francoischalifour: Can we settle on a single name for consistency everywhere? The prop is named `initialState` so I would name the variable this way too.

packages/recommend-core/src/types/InitialResults.ts
```diff
@@ -0,0 +1,5 @@
+import { ProductRecord } from './ProductRecord';
+
+export type InitialResults<TObject> = {
```
francoischalifour: Related to the naming again, can we name this type the same as the prop name? (`InitialState`)

packages/recommend-react/src/__tests__/useFrequentlyBoughtTogether.test.ts
francoischalifour: I don't understand why this is needed.

Should we rather call `rerender()` and then `waitFor()` the `expect`? (Or perhaps we need to wrap this `rerender()` in `act()`.)

packages/recommend-react/src/__tests__/useRecommendations.test.ts
```diff
@@ -41,9 +42,108 @@ describe('useRecommendations', () => {
...
+    expect(recommendClient.getRecommendations).toHaveBeenCalledTimes(0);
+
+    // we force a delay in order to simulate a prop change
+    // (that is a dependency prop in the useEffect from the hook we are testing)
+    // so that we can test that a network call is triggered
+    await forceDelay(1000);
+
+    rerender({ indexName: 'test1' });
+
+    expect(recommendClient.getRecommendations).toHaveBeenCalledTimes(1);
```
francoischalifour, line 183: We could assert that it was called with the correct `indexName`.

packages/recommend-react/src/useRecommendations.ts
```diff
@@ -19,35 +20,48 @@ export function useRecommendations<TObject>({
...
   transformItems: userTransformItems,
-}: GetRecommendationsProps<TObject>) {
-  const [result, setResult] = useState<GetRecommendationsResult<TObject>>({
-    recommendations: [],
-  });
+  initialState: userInitialState,
+}: GetRecommendationsProps<TObject> & {
+  initialState?: InitialResults<TObject>;
+}) {
+  const isFirstRenderRef = useRef(false);
```
francoischalifour: Technically we're in the first render the first time this executes, so we should reverse the booleans.

```suggestion
  const isFirstRenderRef = useRef(true);
```

packages/recommend-react/src/useRecommendations.ts
```diff
@@ -19,35 +20,48 @@ export function useRecommendations<TObject>({
...
+
   const { status, setStatus } = useStatus('loading');
   const objectIDs = useStableValue(userObjectIDs);
   const transformItems = useStableValue(userTransformItems);
+  const initialState = useStableValue(userInitialState);
```
francoischalifour: We shouldn't need the newly introduced variables with something like that:

```tsx
const initialState = useStableValue<GetRecommendationsResult<TObject>>({
  recommendations: [],
  ...userInitialState,
});
const [result, setResult] = useState<GetRecommendationsResult<TObject>>(
  initialState
);
```

packages/recommend-react/src/useRelatedProducts.ts
```diff
@@ -18,34 +19,47 @@ export function useRelatedProducts<TObject>({
...
+        recommendClient,
+        threshold,
+        transformItems,
+      }).then((response) => {
+        setRecommendationsResult(response);
+        setStatus('idle');
+      });
+    } else {
+      isFirstRenderRef.current = true;
+    }
```
francoischalifour: I _think_ there are some broken behaviors in the way that the ref is reset. It needs to be reset either way after this effect.

I also think that having an intermediate `shouldFetch` variable is helpful here to make the intent clear why we have this guard.

```suggestion
    const shouldFetch = !initialState || !isFirstRenderRef.current;

    if (shouldFetch) {
      setStatus('loading');
      getRecommendations({
        fallbackParameters,
        indexName,
        maxRecommendations,
        model,
        objectIDs,
        queryParameters,
        recommendClient,
        threshold,
        transformItems,
      }).then((response) => {
        setResult(response);
        setStatus('idle');
      });
    }

    isFirstRenderRef.current = false;
```

packages/recommend-react/src/__tests__/useFrequentlyBoughtTogether.test.ts
```diff
@@ -37,9 +37,99 @@ describe('useFrequentlyBoughtTogether', () => {
...
+        }),
+      {
+        wrapper: React.StrictMode,
+      }
+    );
+
+    await waitFor(() => {
+      expect(result.current.recommendations).toEqual(
+        initialState.recommendations
+      );
```
francoischalifour: We could check the whole `result.current` object to make sure that `status` is also correct.

packages/recommend-react/src/useRecommendations.ts
```diff
@@ -19,35 +20,51 @@ export function useRecommendations<TObject>({
...
   const objectIDs = useStableValue(userObjectIDs);
   const transformItems = useStableValue(userTransformItems);
   const queryParameters = useStableValue(userQueryParameters);
   const fallbackParameters = useStableValue(userFallbackParameters);
 
+  const initialState = useStableValue<GetRecommendationsResult<TObject>>({
+    recommendations: [],
+    ...userInitialState,
+  });
+  const [recommendationsResult, setRecommendationsResult] = useState<
```
francoischalifour: We can rollback to the previous naming with `[result, setResult]`.

francoischalifour (commented):
Do you plan to also add this feature to `@algolia/recommend-js`?

packages/recommend-react/src/__tests__/useFrequentlyBoughtTogether.test.ts
marialungu: I don't think is necessary as this feature will be used for React SSR 🤔

packages/recommend-react/src/useRecommendations.ts
marialungu: I was thinking that the naming now is a bit more descriptive for what it represents 🤔

packages/recommend-react/src/__tests__/useFrequentlyBoughtTogether.test.ts
francoischalifour: It is because it actually reveals a bug right now: https://github.com/algolia/recommend/blob/04f0380be41374298eb2e186c4e82dc024cb0f99/packages/recommend-react/src/useFrequentlyBoughtTogether.ts#L27

The status is `loading` although it should be `idle` when `initialState` is provided.

packages/recommend-react/src/useRecommendations.ts
francoischalifour: In this context we know that we're talking about recommendations.

packages/recommend-react/src/__tests__/useFrequentlyBoughtTogether.test.ts
marialungu: ah right, will update the status accordingly

marialungu: there's a de-sync in the comments: 
"I don't think is necessary as this feature will be used for React SSR 🤔" -> in response to "Do you plan to also add this feature to @algolia/recommend-js?"

marialungu: update: https://github.com/algolia/recommend/pull/109/commits/64f060fd04ce1c559d047d98f4d5fa10e827f1c1

packages/recommend-react/src/useRecommendations.ts
marialungu: ✅ : https://github.com/algolia/recommend/pull/109/commits/0a9ad61eccb9f34cc04df3ca5981bc596508ff75

packages/recommend-react/src/useRelatedProducts.ts
marialungu: ✅

packages/recommend-react/src/useRecommendations.ts
marialungu: ✅

packages/recommend-react/src/useRecommendations.ts
marialungu: ✅

packages/recommend-core/src/types/InitialResults.ts
marialungu: ✅

examples/server-demo/src/server.js
marialungu: ✅

examples/server-demo/src/server.js
marialungu: ✅

packages/recommend-react/src/__tests__/useFrequentlyBoughtTogether.test.ts
marialungu: ✅

packages/recommend-react/src/useFrequentlyBoughtTogether.ts
```diff
@@ -17,32 +18,51 @@ export function useFrequentlyBoughtTogether<TObject>({
...
+        queryParameters,
+        recommendClient,
+        threshold,
+        transformItems,
+      }).then((response) => {
+        setResult(response);
+        setStatus('idle');
+      });
+    } else {
       setStatus('idle');
```
francoischalifour: We don't need to trigger a state change (and therefore a render) for this. We can directly initialize it in the state declaration:

```tsx
const { status, setStatus } = useStatus(userInitialState ? 'idle' : 'loading');
```

marialungu: keep forgetting to take re-renders into account, it's like it's my first day in React 🤦🏻‍♀️ 
✅ : https://github.com/algolia/recommend/pull/109/commits/bce36b23872ceeeb5ce80462e96ba983a9ff0760

packages/recommend-react/src/useRelatedProducts.ts
```diff
@@ -18,34 +19,55 @@ export function useRelatedProducts<TObject>({
...
+        queryParameters,
+        recommendClient,
+        threshold,
+        transformItems,
+      }).then((response) => {
+        setResult(response);
+        setStatus('idle');
+      });
+    } else {
       setStatus('idle');
```
francoischalifour: This is still remaining :(

packages/recommend-react/src/useTrendingFacets.ts
```diff
@@ -18,33 +19,54 @@ export function useTrendingFacets<TObject>({
...
+        maxRecommendations,
+        queryParameters,
+        threshold,
+        facetName,
+      }).then((response) => {
+        setResult(response);
+        setStatus('idle');
+      });
+    } else {
       setStatus('idle');
```
francoischalifour: Same here.

packages/recommend-react/src/useTrendingItems.ts
```diff
@@ -19,34 +20,55 @@ export function useTrendingItems<TObject>({
...
+        queryParameters,
+        threshold,
+        facetName,
+        facetValue,
+      }).then((response) => {
+        setResult(response);
+        setStatus('idle');
+      });
+    } else {
       setStatus('idle');
```
francoischalifour: Same here.

francoischalifour (commented):
We're almost good to go.

Would you like to have a review @sarahdayan @Haroenv?

examples/server-demo/src/App.js
```diff
@@ -0,0 +1,70 @@
...
+          recommendClient={recommendClient}
+          indexName={indexName}
+          objectIDs={['M0E20000000DWIV']}
+          itemComponent={ProductItem}
+          maxRecommendations={5}
+          view={GridView}
+          translations={{
+            title: 'Frequently bought together',
+          }}
+          initialState={initialState}
```
Haroenv: this is actually the initial results, not the initial state, right? Not sure if the name is chosen correctly

packages/recommend-react/src/useTrendingItems.ts
marialungu: ✅

packages/recommend-react/src/useTrendingFacets.ts
marialungu: ✅

packages/recommend-react/src/useRelatedProducts.ts
marialungu: moved too fast for webstorm to comprehend the file changes😅

marialungu: ✅

Haroenv (commented):
comments left by Francois are indeed valid, and I'm not sure if this should be called initialState or initialResults (or maybe just results?) Is a prop the right place to inject initial results? should that maybe be done to the client (but it seems that isn't available)?

examples/server-demo/src/App.js
marialungu: yes, those would be the initialResults for now. they are injected as a prop, we want to use this prop moving fwd for SSR, injecting these initialResults from the server (so technically it would be an initialState of the app at first load? 🤔 )

Haroenv: It would still be the initial results when it's using real server side rendering no? the state is which parameters get sent, and the results are what is displayed. Especially this code for me indicates it's only results: 

https://github.com/algolia/recommend/pull/109/files#diff-22a64e6cc4de6dd486bbe8a1b8b0a65d2ad9f2564e01becb38de38d3a5af11e2R38-R40

Haroenv (changes requested):
it's obviously your codebase so your call at the end of the day, but I think it should be results, not state

francoischalifour:
@Haroenv Perhaps even `initialRecommendations` since users manipulate `useRecommendations().recommendations`?

I also think that it would make more sense to call this "state" once we have a `<Recommend>` provider that orchestrate these states:

```jsx
<Recommend recommendClient={recommendClient} initialState={initialState}>
  <RelatedProducts />
  <FrequentlyBoughtTogether />
</Recommend>
```

packages/recommend-react/src/__tests__/useRecommendations.test.ts
marialungu: ✅ : https://github.com/algolia/recommend/pull/109/commits/466eafeb0672ced162687b66cc614c70d4b792f3

packages/recommend-react/src/__tests__/useFrequentlyBoughtTogether.test.ts
```diff
@@ -83,4 +81,102 @@ describe('useFrequentlyBoughtTogether', () => {
...
+    expect(recommendClient.getFrequentlyBoughtTogether).toHaveBeenCalledWith([
+      {
+        indexName: 'test1',
+        maxRecommendations: undefined,
+        objectID: 'testing',
+        queryParameters: { facetFilters: ['test'] },
+        threshold: 0,
+      },
+    ]);
+  });
```
francoischalifour: We can make these tests more resilient by asserting just what this test is responsible for. The rest should already be tested elsewhere.

```suggestion
    expect(recommendClient.getFrequentlyBoughtTogether).toHaveBeenCalledWith([
      expect.objectContaining({
        indexName: 'test1',
      }),
    ]);
  });
```

packages/recommend-core/src/types/InitialState.ts
```diff
@@ -0,0 +1,5 @@
+import { ProductRecord } from './ProductRecord';
+
+export type InitialState<TObject> = {
```
francoischalifour: See [Slack thread](https://algolia.slack.com/archives/C03QWAMAN6P/p1668617762173699?thread_ts=1668597136.031949&cid=C03QWAMAN6P). There's still ongoing discussions about this naming before moving forward.

packages/recommend-react/src/__tests__/useFrequentlyBoughtTogether.test.ts
marialungu: ✅ : https://github.com/algolia/recommend/pull/109/commits/55280c4a963e58e3a535634c95961ba4342c01c9

packages/recommend-core/src/types/InitialState.ts
marialungu: ✅ : https://github.com/algolia/recommend/pull/109/commits/ab1e664d713cbde4416ec8d201e5f97d601e6460

francoischalifour: If the object structure is `{ status, recommendations }`, then `initialResult` is more correct than `initialRecommendations`.

marialungu: the structure only has recommendations:
```
export type InitialRecommendations<TObject> = {
  recommendations: Array<ProductRecord<TObject>>;
};
```

francoischalifour: Yes, it _has_ recommendations, it _is not_ recommendations.

marialungu: ✅ : https://github.com/algolia/recommend/pull/109/commits/dd46e702b525bf33cefd88a3222e5214adbea3ba

packages/recommend-react/src/useFrequentlyBoughtTogether.ts
```diff
@@ -21,14 +22,26 @@ export function useFrequentlyBoughtTogether<TObject>({
   recommendClient,
   threshold,
   transformItems: userTransformItems,
-}: UseFrequentlyBoughtTogetherProps<TObject>) {
-  const [result, setResult] = useState<GetRecommendationsResult<TObject>>({
-    recommendations: [],
-  });
-  const { status, setStatus } = useStatus('loading');
+  initialResults: userInitialResults,
```
Haroenv: does this still need to be renamed, it could be this right (with the usage changed too)?

```suggestion
  initialResults,
```

packages/recommend-react/src/useRecommendations.ts
```diff
@@ -21,15 +22,27 @@ export function useRecommendations<TObject>({
   recommendClient,
   threshold,
   transformItems: userTransformItems,
-}: UseRecommendationsProps<TObject>) {
-  const [result, setResult] = useState<GetRecommendationsResult<TObject>>({
-    recommendations: [],
-  });
-  const { status, setStatus } = useStatus('loading');
+  initialResults: userInitialResults,
```
Haroenv: same question

packages/recommend-react/src/useRelatedProducts.ts
```diff
@@ -20,15 +21,27 @@ export function useRelatedProducts<TObject>({
   recommendClient,
   threshold,
   transformItems: userTransformItems,
-}: UseRelatedProductsProps<TObject>) {
-  const [result, setResult] = useState<GetRecommendationsResult<TObject>>({
-    recommendations: [],
-  });
-  const { status, setStatus } = useStatus('loading');
+  initialResults: userInitialResults,
```
Haroenv: and here too of course

packages/recommend-react/src/useTrendingFacets.ts
```diff
@@ -20,14 +21,26 @@ export function useTrendingFacets<TObject>({
   threshold,
   transformItems: userTransformItems,
   facetName,
-}: UseTrendingFacetsProps<TObject>) {
-  const [result, setResult] = useState<GetTrendingFacetsResult<TObject>>({
-    recommendations: [],
-  });
-  const { status, setStatus } = useStatus('loading');
+  initialResults: userInitialResults,
```
Haroenv: idem

packages/recommend-react/src/useTrendingItems.ts
```diff
@@ -21,14 +22,26 @@ export function useTrendingItems<TObject>({
   transformItems: userTransformItems,
   facetName,
   facetValue,
-}: UseTrendingItemsProps<TObject>) {
-  const [result, setResult] = useState<GetTrendingItemsResult<TObject>>({
-    recommendations: [],
-  });
-  const { status, setStatus } = useStatus('loading');
+  initialResults: userInitialResults,
```
Haroenv: idem

examples/server-demo/package.json
```diff
@@ -0,0 +1,29 @@
...
+    "@algolia/client-search": "4.10.5",
+    "@algolia/recommend": "4.13.0",
+    "@algolia/recommend-react": "1.8.0",
+    "algoliasearch": "4.10.5",
+    "react": "17.0.2",
+    "react-dom": "17.0.2",
+    "express": "4.17.1"
+  },
+  "devDependencies": {
+    "@babel/core": "7.4.5",
```
Haroenv, line 20: if we use the same versions as used in the root already it avoids duplicating in yarn.lock

marialungu: we don't have @babel/core in the root

packages/recommend-react/src/useFrequentlyBoughtTogether.ts
marialungu: ✅ : https://github.com/algolia/recommend/pull/109/commits/fe195d38f2074d588c9e2549d77260eb3360bf3c

got a const here that is used to stabilize the initialResults and initialize the useState here: 
```
const initialResults = useStableValue<GetRecommendationsResult<TObject>>({
    recommendations: [],
    ...userInitialResults,
  });
  const [result, setResult] = useState<GetRecommendationsResult<TObject>>(
    initialResults
  );
  ```
  
  but i also need the initialResults from the user here, hence the renaming:
  ```
  const shouldFetch = !userInitialResults || !isFirstRenderRef.current;
  ```

File: examples/server-demo/package.json (hunk 37 above)
Comment (Haroenv) [reply examples/server-demo/package.json:20]: I just see that it added some things to the lock file that could be harmonised / made compatible if this was ^7.4.5 or the exact version used somewhere else in the project. I guess it's not a big problem though

**Review by Haroenv — APPROVED** [review]

File: packages/recommend-react/src/useFrequentlyBoughtTogether.ts [hunk 38]
```diff
@@ -34,12 +34,12 @@ export function useFrequentlyBoughtTogether<TObject>({
...
   const queryParameters = useStableValue(userQueryParameters);
 
-  const initialState = useStableValue<GetRecommendationsResult<TObject>>({
+  const initialResults = useStableValue<GetRecommendationsResult<TObject>>({
     recommendations: [],
     ...userInitialResults,
   });
   const [result, setResult] = useState<GetRecommendationsResult<TObject>>(
-    initialState
+    initialResults
```
francoischalifour: There's still a vocabulary mismatch. Here we're storing results (plural) in result.

Should we go with the singular form?

marialungu: ah, it is a bit weird 🤔 
something like we have `initialReuslts` cuz they represent the intial recommendation**s** provided (as of now) by the user and we have `result` as in the API result .... wdyt?

Haroenv: initialResult everywhere would make sense to me, it's one result object which has multiple recommendations inside

francoischalifour: I agree with @Haroenv.

marialungu: ✅ : https://github.com/algolia/recommend/pull/109/commits/adffbda5888120c89018516ff360980a7fc493a1

francoischalifour (approved):

aleEspinosaM:
🥳 🥳 🥳 🥳 🥳 🥳

Issue #62, opened 2022-01-20
`transformItems` does not provide a reliable data structure for `items`


dhayab:
## Description

When using `useFrequentlyBoughtTogether` or `useRelatedProducts`, the `transformItems` property should allow me to retrieve the list of items, transform it and return a new list.

It appears the list of items (`items`) is not always an Array, making it impossible to reliably iterate over it or use any Array function (`map`, `filter`, etc.).

## Reproduction

In this sandbox, the Frequently Bought Together component is wrapped in a `connectStateResults` connector, allowing me to retrieve the ID of the first Algolia hit.

[**Preview →**](https://codesandbox.io/s/trusting-violet-fqb7q?file=/src/App.js)

**Steps**

Just open the console and see subsequents calls to `transformItems` providing a wrong structure as `items`.

## Expected behavior

`items` is always an Array.

## Environment

- OS: macOS
- Browser: Chrome
- Library version:
  - @algolia/recommend: 4.12.0
  - @algolia/recommend-react: 1.1.0
  - react-instantsearch-dom: 6.0.0

Haroenv:
relevant code: https://github.com/algolia/recommend/blob/e0318420a12100658e9d8a04752f6464d874e227/packages/recommend-core/src/getFrequentlyBoughtTogether.ts#L39

is it hits that can be null or something else?

dhayab:
From what I see in the console, hits can be a reference to `transformItems()`.

evanwinter:
Any progress on this? I'm running into this too.

```
const { recommendations } = useRelatedProducts({
    recommendClient,
    indexName,
    objectIDs,
    queryParameters: { filters },
    transformItems: (items) => {
        console.log(items);
        return items;
    },
});
```

<img width="464" alt="Screen Shot 2022-10-13 at 11 32 45 AM" src="https://user-images.githubusercontent.com/19195468/195654237-b627a163-8a68-4a7d-b096-b3323e131569.png">


On the first cycle, `items` is an Array but every cycle after that it looks like it's a reference to `transformItems`

francoischalifour:
This happens because we try to stabilize the `transformItems` prop: https://github.com/algolia/recommend/blob/d8e63758684241ea24d2abb5465e47b38cb32225/packages/recommend-react/src/useTrendingItems.ts#L27

[...]

PR #110, opened 2022-10-24
fix(react): correctly stabilize `transformItems` function

marialungu:
Correctly stabilize `transformItems` function from the hooks, as reported in this [issue](https://github.com/algolia/recommend/issues/62)

### DIFF
--- a/packages/recommend-react/src/__tests__/useFrequentlyBoughtTogether.test.ts
+++ b/packages/recommend-react/src/__tests__/useFrequentlyBoughtTogether.test.ts
@@ -8,6 +8,7 @@ import {
   hit,
 } from '../../../../test/utils/createRecommendClient';
 import { useFrequentlyBoughtTogether } from '../useFrequentlyBoughtTogether';
+import {getItemName, getItemPrice} from "../../../../test/utils";
 
 function createMockedRecommendClient() {
   const recommendClient = createRecommendClient({
@@ -26,7 +27,7 @@ function createMockedRecommendClient() {
 }
 
 describe('useFrequentlyBoughtTogether', () => {
-  test('returns FBT recommendations', async () => {
+  test.skip('returns FBT recommendations', async () => {
     const { recommendClient } = createMockedRecommendClient();
 
     const { result } = renderHook(() =>
@@ -45,7 +46,7 @@ describe('useFrequentlyBoughtTogether', () => {
     });
   });
 
-  test('assures that the transformItems function always returns an array', async () => {
+  test.skip('assures that the transformItems function always returns an array', async () => {
     const { recommendClient } = createMockedRecommendClient();
 
     const { result, rerender } = renderHook(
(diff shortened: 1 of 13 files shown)


packages/recommend-react/src/useFrequentlyBoughtTogether.ts
```diff
@@ -23,9 +23,11 @@ export function useFrequentlyBoughtTogether<TObject>({
   });
   const { status, setStatus } = useStatus('loading');
   const objectIDs = useStableValue(userObjectIDs);
-  const transformItems = useStableValue(userTransformItems);
   const queryParameters = useStableValue(userQueryParameters);
 
+  const transformItemsRef = useRef(userTransformItems);
+  const transformItems = transformItemsRef.current;
```
francoischalifour: Can you add some tests by updating the `transformItems` prop? You'll realize this doesn't work.

You need to reassign `transformItemsRef.current` to the `transformItems` prop in an effect, and then use `transformItemsRef.current` in the existing effect.

francoischalifour (changes requested):

packages/recommend-react/src/__tests__/useFrequentlyBoughtTogether.test.ts
```diff
@@ -24,22 +26,62 @@ function createMockedRecommendClient() {
...
+      }
+    );
+    await waitFor(() => {
+      expect(result.current.recommendations).toEqual([
+        'Landoh 4-Pocket Jumpsuit',
+      ]);
+    });
+
+    act(() => {
+      rerender();
```
francoischalifour: This tests that the weird array behavior is gone, but it doesn't test that the feature actually works.

Here, we want to rerender with another `transformItems` reference, and to make sure that the items have been transformed with this latest function reference.

packages/recommend-react/src/useFrequentlyBoughtTogether.ts
```diff
@@ -50,7 +54,7 @@ export function useFrequentlyBoughtTogether<TObject>({
     recommendClient,
     setStatus,
     threshold,
-    transformItems,
+    transformItemsRef,
```
francoischalifour: This shouldn't be needed?

packages/recommend-react/src/useFrequentlyBoughtTogether.ts
```diff
@@ -23,9 +23,13 @@ export function useFrequentlyBoughtTogether<TObject>({
   });
   const { status, setStatus } = useStatus('loading');
   const objectIDs = useStableValue(userObjectIDs);
-  const transformItems = useStableValue(userTransformItems);
   const queryParameters = useStableValue(userQueryParameters);
 
+  const transformItemsRef = useRef(userTransformItems);
+  useEffect(() => {
+    transformItemsRef.current = userTransformItems;
+  }, [userTransformItems]);
```
francoischalifour, line 33: So that the `useAlgoliaAgent()` call doesn't get missed/lost, perhaps we can move this code below it.

packages/recommend-react/src/__tests__/useFrequentlyBoughtTogether.test.ts
```diff
@@ -24,22 +27,100 @@ function createMockedRecommendClient() {
...
-        transformItems: (items) => items,
-      });
+      })
+    );
+    await waitFor(() => {
+      expect(result.current.recommendations).toEqual([hit]);
+    });
+  });
+
+  test('assures that the transformItems function always returns an array', async () => {
```
francoischalifour: We can remove this test now, it's covered with the one below.

francoischalifour:
@Haroenv @sarahdayan @dhayab Are you happy with this function stability solution?

It automatically "caches" the `transformItems` function in a ref so that users can pass unstable function references.

However, when the `transformItems` function reference changes, it doesn't re-trigger a render. It's only on the next non-function prop change that the effect re-executes.

Haroenv:
If you're purposely changing the transformItems function you would expect it to immediately be evaluated, however if you simply pass an unstable function you wouldn't expect that to have a side effect. As those behaviours are incompatible, the most correct is indeed probably what you're suggesting here; i.e. stabilising the function, but not causing any side effects through any changes that would happen

packages/recommend-react/src/__tests__/useFrequentlyBoughtTogether.test.ts
marialungu: ✅ : https://github.com/algolia/recommend/pull/110/commits/9214aa08f43745d5dda1736e1fc1dd4118f5e038

packages/recommend-react/src/useFrequentlyBoughtTogether.ts
marialungu: ✅ : https://github.com/algolia/recommend/pull/110/commits/9214aa08f43745d5dda1736e1fc1dd4118f5e038

packages/recommend-react/src/useFrequentlyBoughtTogether.ts
marialungu: ✅

francoischalifour (approved):

PR #112, opened 2022-11-07
feat(react): export props types

francoischalifour:
The components and Hooks props were not exported so it wasn't simple to re-create higher-level components reusing these props.

In the other libraries, we make sure to export the types that are close to the developer usage.


Haroenv (approved):
makes sense, even if it can easily be retrieved via ComponentProps and Arguments

Haroenv:
in the future api-explorer can be used to enforce this

marialungu (approved):

PR #118, opened 2022-12-15
fix(recommend): remove unused trendingFacets props (RECO-1173)

marialungu:
Remove `fallbackParameters` and `queryParameters` given that the recommend API does not support these props

### DIFF
--- a/packages/recommend-core/src/getTrendingFacets.ts
+++ b/packages/recommend-core/src/getTrendingFacets.ts
@@ -29,18 +29,14 @@ export type GetTrendingFacetsProps<TObject> = TrendingFacetsProps<TObject> &
 export function getTrendingFacets<TObject>({
   recommendClient,
   transformItems = (x) => x,
-  fallbackParameters,
   indexName,
   maxRecommendations,
-  queryParameters,
   threshold,
   facetName,
 }: GetTrendingFacetsProps<TObject>) {
   const query = {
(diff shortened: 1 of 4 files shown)


packages/recommend-core/src/getTrendingFacets.ts
```diff
@@ -29,18 +29,14 @@ export type GetTrendingFacetsProps<TObject> = TrendingFacetsProps<TObject> &
 export function getTrendingFacets<TObject>({
   recommendClient,
   transformItems = (x) => x,
-  fallbackParameters,
   indexName,
   maxRecommendations,
-  queryParameters,
   threshold,
   facetName,
 }: GetTrendingFacetsProps<TObject>) {
```
francoischalifour, line 36: In the API client, the `TrendingFacetsQuery` type also needs to be adapted, because it still contains `queryParameters` and `fallbackParameters`.

francoischalifour (approved):

packages/recommend-core/src/getTrendingFacets.ts
marialungu: yep, prepping a PR right now

jojva (approved):
LGTM :+1:

jojva (approved):

francoischalifour:
@marialungu Did you figure out why the bundle size increased since this PR removes code?

PR #123, opened 2023-02-10
Expose queryID through frequentlyBoughtTogether

jadenguitarman:


marialungu:
Hi!

Thank you for your suggestion, we opened this [PR](https://github.com/algolia/recommend/pull/124) with a possible  implementation of this feature. 
We are currently investigating it further, in order to asses the impact of exposing this `queryID`. 
We will let you know when the full solution is available.

Thank you

PR #130, opened 2023-04-17
feat: update react playground demo

raed667:
Update the React demo to use `react-router-dom` 

- Split the logic into separate Home/Product pages for easier reading. 
- Update `search-insights` to remove some 422 errors.
- ~Update demo to React 18~ reverted due to some issue with the monorepo, to be investigated 

Should be pretty much iso-functional with existing demo.


aleEspinosaM (approved):
LGTM!

PR #131, opened 2023-04-17
feat: implement Recommend Batch Context (RECO-961)

raed667:
# Implement Recommend (Batch) Context

## 🤔 What

Create a Recommend React Provider and enable batching. 

More details at [RECO-1400] and `RFC - recommend-react - Recommend Provider & Request batching`

## 🤷‍♂️ Why

Recommend REST API allows for query Batching, we want to make this feature more accessible to users who use the React UI library.

## 🔍 How

Users who want to take advantage of query Batching can wrap their components with a `<Recommend>` provider. 

Each Recommend UI component will register its query parameters with the parent context and requests will be automatically batched. 

To opt a specific component out of batching (using a different index for example) the user can pass a `recommendClient` instance with the appropriate parameters to that particular component.

## 🧪 Testing

Batching can be tested in action by running the React demo project under examples 

```sh
$ yarn playground:start
```

Check the network tab and see how different recommend queries are batched with a single HTTP request.

[RECO-1400]: https://algolia.atlassian.net/browse/RECO-1400?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/recommend-react/src/RecommendContext.tsx
```diff
@@ -0,0 +1,207 @@
...
+    case 'request_success': {
+      return { ...state, status: 'idle', isDirty: null };
+    }
+
+    default:
+      return state;
+  }
+};
+
+export function RecommendProvider<TObject>({
```
francoischalifour: Why not call this `<Recommend>` like for `<InstantSearch>` and `<Predict>`?

packages/recommend-react/src/RecommendContext.tsx
```diff
@@ -0,0 +1,207 @@
...
+}: RecommendProviderProps) {
+  const [state, dispatch] = React.useReducer(reducer, {
+    isDirty: null,
+    cache: {},
+    widgets: [],
+    recommendClient,
+  });
+
+  React.useEffect(() => {
+    (async () => {
```
francoischalifour: We've been avoiding async functions in our libraries because the tooling doesn't transpile them down to our browser target IIRC.

packages/recommend-react/src/RecommendContext.tsx
```diff
@@ -0,0 +1,207 @@
...
+    []
+  );
+
+  return (
+    <RecommendContext.Provider
+      value={{
+        register,
+        recommendClient,
+        hasProvider: true,
+      }}
```
francoischalifour: In our libraries we likely always want to stabilize context values to minimize renders in our customers' apps.

packages/recommend-react/src/RecommendContext.tsx
```diff
@@ -0,0 +1,207 @@
...
+  widgets: Array<RecommendWidget<TObject>>;
+  recommendClient: RecommendClient | null;
+};
+
+type Action<TObject> =
+  | { type: 'register'; widget: Omit<RecommendWidget<TObject>, 'param'> }
+  | { type: 'unregister'; key: string }
+  | { type: 'request_success' };
+
+export const RecommendContext = React.createContext<
```
francoischalifour, line 33: For better debugging experience in the DevTools, we can add a `displayName` to the context in development. See [how it's done in Predict](https://github.com/algolia/predict/blob/c5040758c6807e6dd5d5b10d1e7d7f463514b353/packages/predict-react/src/lib/PredictContext.ts#L65-L67).

packages/recommend-react/src/index.ts
```diff
@@ -7,3 +7,4 @@ export * from './useRecommendations';
 export * from './useRelatedProducts';
 export * from './useTrendingFacets';
 export * from './useTrendingItems';
+export * from './RecommendContext';
```
francoischalifour: This export means that `RecommendContext` and `useRecommendContext` are also exposed to users, but they should probably remain internals.

Having an explicit `Recommend.tsx` file (see my suggestion above renaming `RecommendProvider` to `Recommend`) would make the entry point easier to find for contributors, and likely reduce the risk of exporting more than needed.

packages/recommend-react/src/RecommendContext.tsx
```diff
@@ -0,0 +1,207 @@
...
+
+  const register = React.useCallback(
+    (widget: Omit<RecommendWidget<TObject>, 'param'>) => {
+      dispatch({ type: 'register', widget });
+      return (key: string) => {
+        dispatch({ type: 'unregister', key });
+      };
+    },
+    []
+  );
```
francoischalifour: We could probably remove some `@ts-ignore` if we could call this function with a generic: `register<RecommendationType>(...)`.

packages/recommend-react/src/useFrequentlyBoughtTogether.ts
```diff
@@ -29,35 +32,98 @@ export function useFrequentlyBoughtTogether<TObject>({
...
   const queryParameters = useStableValue(userQueryParameters);
 
-  useAlgoliaAgent({ recommendClient });
+  const {
+    recommendClient: recommendClientFromContext,
+    hasProvider,
+    register,
+  } = useRecommendContext();
+
+  const { client, isContextClient } = pickRecommendClient(
```
francoischalifour: What is `isContextClient` for? Isn't the provider check enough?

packages/recommend-react/src/RecommendContext.tsx
```diff
@@ -0,0 +1,207 @@
...
+  onRequest: () => void;
+  key: string;
+  param: GetParametersResult;
+};
+
+type RecommendContextType<TObject> = {
+  register: (
+    widget: Omit<RecommendWidget<TObject>, 'param'>
+  ) => (key: string) => void;
+  hasProvider: boolean;
```
francoischalifour, line 24: Can't we rely on passing `null` as the initial context?

packages/recommend-react/src/useFrequentlyBoughtTogether.ts
raed667: It is for when we have both `recommendClient` instances passed to the provider and the child component. 

If a child component also has `recommendClient` we privilege that one as an override instead of the `recommendClient` passed to the provider.

packages/recommend-react/src/useFrequentlyBoughtTogether.ts
```diff
@@ -29,35 +32,98 @@ export function useFrequentlyBoughtTogether<TObject>({
...
       objectIDs,
       queryParameters,
-      recommendClient,
       threshold,
       transformItems: transformItemsRef.current,
-    }).then((response) => {
-      setResult(response);
-      setStatus('idle');
-    });
+    };
```
francoischalifour: Did you mean to reuse this params in the calls below as well?

francoischalifour (commented):
I reviewed the library aspects but not yet the batching logic.

Sorry for the noise I was using another review tool and meant to send a single review.

packages/recommend-react/src/RecommendContext.tsx
raed667: we can but it would make destructuring in the components a 2 lines operation instead of the straightforward way we do it today. 

I don't see the benefit of null in this case?

packages/recommend-react/src/useFrequentlyBoughtTogether.ts
raed667: Indeed !

packages/recommend-react/src/RecommendContext.tsx
raed667: register is using `React.useCallback` and `hasProvider` is a constant, do you propose we stabilize `recommendClient` as such ? 

```js
  const recommendClient = useStableValue(userRecommendClient);
```

packages/recommend-react/src/RecommendContext.tsx
raed667: My TypeScript kung-fu is apparently not strong enough for this, I have tried a couple of approaches of passing the expected return type from the hook to the context but i can't seem to crack it :(

packages/recommend-react/src/Recommend.tsx
```diff
@@ -0,0 +1,235 @@
...
+        };
+      }
+
+      return state;
+    }
+
+    case 'unregister': {
+      const { key } = action;
+      state.widgets = state.widgets.filter((w) => w.key !== key);
+      return state;
```
francoischalifour: We're supposed to return a new state object in the reducer, not mutate the previous one. See the pitfall section in this [React doc](https://react.dev/reference/react/useReducer#writing-the-reducer-function).

packages/recommend-react/src/index.ts
```diff
@@ -7,3 +7,4 @@ export * from './useRecommendations';
 export * from './useRelatedProducts';
 export * from './useTrendingFacets';
 export * from './useTrendingItems';
+export { Recommend } from './Recommend';
```
francoischalifour: We also want to export `RecommendProps`.

packages/recommend-core/src/getBatchRecommendations.ts
```diff
@@ -0,0 +1,70 @@
...
+  value: number;
+};
+
+export type GetBatchRecommendations<TObject> = {
+  keys: BatchKeyPair[];
+  queries: RecommendationsQuery[];
+  recommendClient: RecommendClient;
+  transformItems?: (
+    items: Array<ProductRecord<TObject>>
+  ) => Array<ProductRecord<TObject>>;
```
francoischalifour, line 19: We never apply `transformItems` in the batched recommendations?

packages/recommend-react/src/Recommend.tsx
```diff
@@ -0,0 +1,235 @@
...
+  );
+};
+
+export const useRecommendContext = (
+  recommendClient?: RecommendClient | null
+) => {
+  const context = React.useContext(RecommendContext);
+  const { client, isContextClient } = useRecommendClient(recommendClient);
+  return { ...context, client, isContextClient };
+};
```
francoischalifour: Even if we stabilize the context provider value, this return value will always be a new reference because spreading creates a new object. Why aren't `useRecommendContext` and `useRecommendClient` used separately?

packages/recommend-react/src/Recommend.tsx
```diff
@@ -0,0 +1,235 @@
...
+    default:
+      return state;
+  }
+};
+
+export function Recommend<TObject>({
+  recommendClient: userRecommendClient,
+  children,
+}: RecommendProps) {
+  const recommendClient = useStableValue(userRecommendClient);
```
francoischalifour: The contract is that the client must be a stable reference. We state that in the InstantSearch docs. So we don't need to stabilize it ourselves.

packages/recommend-react/src/Recommend.tsx
```diff
@@ -0,0 +1,235 @@
...
+      dispatch({ type: 'request_success' });
+    });
+  }, [state.isDirty, dispatch, state.widgets, state.cache, recommendClient]);
+
+  const register = React.useCallback(
+    (widget: Omit<RecommendWidget<TObject>, 'param'>) => {
+      dispatch({ type: 'register', widget });
+      return (key: string) => {
+        dispatch({ type: 'unregister', key });
+      };
```
francoischalifour: At this point we're already aware of the `key` via `widget.key` so we don't need to pass `key` to unregister.

packages/recommend-react/src/RecommendContext.tsx
francoischalifour: Each context key is stable, but the context value itself isn't (it's a new object at every render). So we can directly use `useMemo()` here.

packages/recommend-react/src/Recommend.tsx
raed667: Good catch 👍

packages/recommend-react/src/Recommend.tsx
raed667: Good catch 👍

packages/recommend-react/src/Recommend.tsx
raed667: Ah yes ! It's an artifact of a previous approach 👍 Good catch

packages/recommend-react/src/Recommend.tsx
raed667: Oh i thought i was being clever by reducing the number of hooks to call, but you're right. I'll put it back in 2 different hooks.

packages/recommend-core/src/getBatchRecommendations.ts
raed667: This is an oversight, I'll need to check with @marialungu about how `transformItems` is meant to be used.

I guess my question here would be is it: 

1- one `transformItems` applied everywhere (as the method signature suggests) ? 
2- each query has its own `transformItems` ?
3- a combination of 1 and 2 ?
4- does it get applied before or after `mapToRecommendations` ? 


Because i'm tempted to just do: 


```js
   if (transformItems) {
      recommendations = transformItems(recommendations);
    }
```

packages/recommend-react/src/__tests__/Recommend.test.tsx
```diff
@@ -0,0 +1,110 @@
...
+    act(() => {
+      render(
+        <Recommend recommendClient={recommendClient}>
+          <RecommendComponent model="related-products" objectId="1234" />
+          <RecommendComponent
+            model="frequently-bought-together"
+            objectId="5678"
+          />
+        </Recommend>
+      );
```
francoischalifour: Can we run the tests in Strict Mode?

packages/recommend-react/src/__tests__/Recommend.test.tsx
```diff
@@ -0,0 +1,110 @@
...
+          onResultSpy(result);
+        },
+      });
+    }, [register, objectId, model]);
+    return (
+      <div>
+        <div>hasProvider:{`${hasProvider}`}</div>
+      </div>
+    );
+  };
```
francoischalifour: What's the point of this test component? It basically replicates the internal logic our the official Recommend components. Any change there would make this test meaningless.

Can we make sure to test our public API to bring confidence that it works in customers' applications?

packages/recommend-core/src/__tests__/getBatchRecommendations.test.ts
```diff
@@ -0,0 +1,119 @@
...
+            objectID: 'objectID',
+          },
+        ],
+        recommendClient,
+      });
+
+      expect(mapByScoreToRecommendationsSpy).toHaveBeenCalledTimes(
+        mapByScore ? 1 : 0
+      );
+      expect(mapToRecommendationsSpy).toHaveReturnedTimes(mapTo ? 1 : 0);
```
francoischalifour: Why test that the internal functions were called and not the actual output? That's an implementation detail that doesn't describe what this test is supposed to do. Reading this test, it's not clear what contract we're testing.

packages/recommend-react/src/useTrendingFacets.ts
```diff
@@ -23,33 +26,76 @@ export function useTrendingFacets<TObject>({
...
-    setStatus('loading');
-    getTrendingFacets({
-      recommendClient,
-      transformItems: transformItemsRef.current,
+    const param: RecommendationsQuery = {
+      model: 'trending-facets',
       indexName,
-      maxRecommendations,
-      threshold,
+      // @ts-expect-error
```
francoischalifour: Can we provide an explanation after each `@ts-expect-error` to express why it's needed and to potentially better understand when it can be removed?

packages/recommend-react/src/__tests__/Recommend.test.tsx
```diff
@@ -0,0 +1,110 @@
...
+      )
+    ),
+  });
+
+  return {
+    recommendClient,
+  };
+}
+
+describe('Recommend', () => {
```
francoischalifour: What I would expect from this test is:

- render only provided components (without Recommend client) and assert that the correct Recommend client method is called
- render provided components (without Recommend client) and detached components (with Recommend client) within the provider and assert that the correct Recommend client methods are called on the correct clients.

Basically, the tests should explain the contract as they're written in the RFC.

packages/recommend-react/src/__tests__/Recommend.test.tsx
raed667: The public API is exposed through the hooks. So that has already been tested. 

Since we don't expose internal logic this was my attempt at capturing and testing the life-cycle of the provider. I can't think of another way.  

But indeed this is is already tested through the individual hooks.

packages/recommend-react/src/useTrendingFacets.ts
raed667: I believe this is the last one left that isn't in a test. 

The problem is that `RecommendationsQuery` from `@algolia/recommend` doesn't allow to pass a `facetName` 

Which I was meaning to ask you or @sarahdayan about. 

Is this an omission from `@algolia/recommend` ? Or should I define a type that will be basically `type RecommendationsQuery & {  facetName?:string } ` ?

francoischalifour: Sarah and I didn't introduce the trending features, Sophie did. It's likely a wrong typing.

raed667: I see, thanks for the clarification! I'll have a look at that type and maybe propose a change to fix the issue.

jest.config.js
```diff
@@ -12,4 +12,6 @@ module.exports = {
     'jest-watch-typeahead/filename',
     'jest-watch-typeahead/testname',
   ],
+  collectCoverage: false,
```
Haroenv: why is coverage set up to be collected but set to false?

packages/recommend-core/src/getBatchRecommendations.ts
```diff
@@ -0,0 +1,83 @@
...
+import { version } from './version';
+
+export type BatchKeyPair = {
+  key: string;
+  value: number;
+};
+
+type Recommendation<TObject> = ProductRecord<TObject>;
+
+// | RecordWithObjectID
```
Haroenv: comment should not be merged

packages/recommend-react/src/FrequentlyBoughtTogether.tsx
```diff
@@ -14,10 +15,10 @@ const UncontrolledFrequentlyBoughtTogether = createFrequentlyBoughtTogetherCompo
...
 );
 
-export type FrequentlyBoughtTogetherProps<
-  TObject
-> = GetFrequentlyBoughtTogetherProps<TObject> &
-  Omit<
+export type FrequentlyBoughtTogetherProps<TObject> = Omit<
+  GetFrequentlyBoughtTogetherProps<TObject>,
+  'recommendClient'
+> & { recommendClient?: RecommendClient } & Omit<
```
Haroenv: as this is used multiple times, you can make a utility:

```ts
type OptionalRecommendClient<T> = Omit<T, 'recommendClient'> & { recommendClient?: T['recommendClient'] }
```

packages/recommend-react/src/Recommend.tsx
```diff
@@ -0,0 +1,226 @@
+import { RecommendClient } from '@algolia/recommend';
```
Haroenv, line 1: modelling the entire batch logic in React means that if you're ever interested in enabling this for regular JS or another framework, you'll need to entirely rewrite the logic.

packages/recommend-react/src/__tests__/useTrendingItems.test.tsx
```diff
@@ -43,6 +43,7 @@ describe('useTrendingItems', () => {
         },
         facetName: 'test4',
         facetValue: 'test3',
+        itemComponent: (item) => <>{item}</>,
```
Haroenv: is this change required? maybe it should be done separately from this PR

packages/recommend-react/src/index.ts
```diff
@@ -7,3 +7,5 @@ export * from './useRecommendations';
...
```
Haroenv: what's the reason of not using `export *`? If there are exported internals, those could be moved to a different file.

packages/recommend-react/src/useFrequentlyBoughtTogether.ts
```diff
@@ -29,35 +31,84 @@ export function useFrequentlyBoughtTogether<TObject>({
...
+    let unregister: Function | undefined;
```
Haroenv: is unregister used somewhere else than in the return of the effect? if not it can be moved to the usage and made const

yarn.lock
```diff
@@ -49,11 +49,23 @@
...
+  integrity sha512-ZkVN7K/JE+qMQbpR6h3gQOGR6yCJpmucSBCmH5YDxnrYbp2CbrVCu0Nr+FGVoWzMJNznj1waShkfQ9awERulLw==
+  dependencies:
+    "@algolia/cache-common" "4.17.2"
+
 "@algolia/cache-common@4.17.0":
   version "4.17.0"
   resolved "https://registry.yarnpkg.com/@algolia/cache-common/-/cache-common-4.17.0.tgz#bc3da15548df585b44d76c55e66b0056a2b3f917"
   integrity sha512-g8mXzkrcUBIPZaulAuqE7xyHhLAYAcF2xSch7d9dABheybaU3U91LjBX6eJTEB7XVhEsgK4Smi27vWtAJRhIKQ==
 
+"@algolia/cache-common@4.17.2":
```
Haroenv: you have two versions of the algolia client in lock now, you can run `npx yarn-deduplicate`

packages/recommend-core/src/getBatchRecommendations.ts
raed667: Good catch !

packages/recommend-react/src/Recommend.tsx
raed667: Yes, as far as we know for now. Batch only makes sense in React and that's the only request we got. Otherwise users can just craft their queries manually.

More details can be found in the RFC

yarn.lock
raed667: Didn't know this tool exist ! Thanks

jest.config.js
raed667: We don't have anywhere to track coverage yet, so I thought I'd enable it in a follow-up PR.

packages/recommend-react/src/__tests__/useTrendingItems.test.tsx
raed667: I can do it in another PR that gets merged before 👍

raed667: This was discussed with @francoischalifour as we don't want to export internal details of how the context works to the end users, while keeping them accessible internally.

I don't think file structure should change here.

packages/recommend-react/src/index.ts
raed667: This was discussed with @francoischalifour as we don't want to export internal details of how the context works to the end users, while keeping them accessible internally.

I don't think file structure should change here.

packages/recommend-react/src/useFrequentlyBoughtTogether.ts
raed667: `unregister` is defined conditionally, how would you move it in this case? 

```js
  useEffect(() => {
    let unregister;
    if(isBatch){
      unregister = register({...})
    }else{
      // do something else
    }
    return ()=>{
      if(unregister){
        unregister()
      }
    }
  }, [])
```

packages/recommend-react/src/index.ts
francoischalifour: My point was that we were exporting the React context and its Hook unpurposefully. I do agree that extracting context-related things to another file would be less error prone: we can keep the `export *` pattern.

raed667: Split `Recommend` and `RecommendContext` files

jest.config.js
Haroenv: this part can also be moved into the separate PR then probably

packages/recommend-react/src/useFrequentlyBoughtTogether.ts
Haroenv: ah the file is too long, I did not catch that the return is in a different level.

However what you can do instead is:

```
  useEffect(() => {
    if(isBatch){
      return register({...})
    }

    // do something else
    return () => {}
  }, [])
```

which simplifies and lowers the levels of conditionality needed.

packages/recommend-react/src/Recommend.tsx
Haroenv: I think it does make sense to only implement this in React for now, but the majority of this code could (and IMHO should) be extracted into a separate purpose-built set of functions that can be used outside of react too. In InstantSearch we've had a significant amount of pain in making parts of the code that only supported InstantSearch.js generic, and I'd recommend that core items don't become part of a specific flavour

francoischalifour: I agree, there's long-term benefits in designing these core parts in an agnostic way. Conceptually, this isn't React specific.

packages/recommend-core/src/getBatchRecommendations.ts
```diff
@@ -0,0 +1,77 @@
...
+    } else {
+      recommendations = mapToRecommendations<ProductRecord<TObject>>({
+        maxRecommendations,
+        hits: splitResult.map((res) => res.hits),
+        nrOfObjs: keyPair.value,
+      });
+    }
+    if (transformItems) {
+      recommendations = transformItems(recommendations);
+    }
```
Haroenv: if transformItems had a default of `x => x` you don't need to make this conditional

packages/recommend-core/src/getBatchRecommendations.ts
```diff
@@ -0,0 +1,77 @@
...
+  recommendClient.addAlgoliaAgent('recommend-core', version);
+
+  const response = await recommendClient.getRecommendations<TObject>(queries);
+
+  let prevChunks = 0;
+  let allChunks = 0;
+
+  const results: Record<string, BatchRecommendations<TObject>> = {};
+
+  keys.forEach((keyPair) => {
```
Haroenv, line 48: fine as a forEach, but reduce fits this just as well

Haroenv: or a map and Object.fromEntries

Haroenv (commented):
all existing comments still valid as well

packages/recommend-react/src/useFrequentlyBoughtTogether.ts
raed667: updated 👍

packages/recommend-core/src/getBatchRecommendations.ts
marialungu: ✅ : https://github.com/algolia/recommend/pull/131/commits/8d42c6f2c2015b36593f587f4086912bb5de66e1

packages/recommend-react/src/__tests__/useTrendingItems.test.tsx
marialungu: ✅ :https://github.com/algolia/recommend/pull/131/commits/2af99ace34588ed75b07db4bc3ef619872a9751e

packages/recommend-react/src/FrequentlyBoughtTogether.tsx
marialungu: ✅ : https://github.com/algolia/recommend/pull/131/commits/2af99ace34588ed75b07db4bc3ef619872a9751e

packages/recommend-react/src/Recommend.tsx
Haroenv: so will you not address this?

raed667: "Batching" already works in plain JS. We can call `client.getRecommendations` with all the queries. 

For me "batching" in the context of this PR only makes sense if the user is working in a declarative context. So I don't see what can be made more generic / ported to plain JS?

marialungu (approved):

packages/recommend-core/src/getBatchRecommendations.ts
raed667: I know this is a bit subjective, but I found this to be more readable than the more clever reduce approach.

PR #132, opened 2023-04-18
fix: Fix JS playground

raed667:
Avoid pragma error on JS playground.

### DIFF
--- a/examples/js-demo/RelatedItem.tsx
+++ b/examples/js-demo/RelatedItem.tsx
@@ -1,3 +1,4 @@
+/** @jsxRuntime classic */
 /** @jsx h */
 import { h } from 'preact';
 
--- a/examples/js-demo/app.tsx
+++ b/examples/js-demo/app.tsx
@@ -1,3 +1,4 @@
+/** @jsxRuntime classic */
 /** @jsx h */
 import { autocomplete, getAlgoliaResults } from '@algolia/autocomplete-js';
 import algoliarecommend from '@algolia/recommend';


examples/js-demo/RelatedItem.tsx
```diff
@@ -1,3 +1,4 @@
+/** @jsxRuntime classic */
```
aleEspinosaM, line 1: why do we need to add this, can you give us a little bit more context, in case we encounter more instances of this error?

raed667: It is due to the way the `recommend-vdom` is setup with `preact` & `preact/hooks`. 
We basically need to tell the tooling (babel) how to interpret JSX syntax and transform it into plain JavaScript on the output. 

`/** @jsx h */` -> Tells Babel to transform JSX into Preact `h()` calls instead of `React.createElement`

By default Babel (after v7.9.0) has an "automatic runtime" where it will import all the functions it will compile the JSX to. Since we're not using React but Preact which sets back the old behavior of allowing us to customize those JSX transformation functions.  

https://preactjs.com/guide/v8/getting-started/
https://babeljs.io/docs/babel-plugin-transform-react-jsx#react-classic-runtime 

PS: Thanks for the good question I had to go back and look some references up!

aleEspinosaM: thanks a lot for the detailed explanation 

approve

aleEspinosaM (approved):

PR #133, opened 2023-04-18
fix: Fix Examples TypeScript issues

raed667:
Examples were previously excluded from type-checking which caused a number of issues.

How to test: 

Run the 3 examples one after the other and check that they work properly. No library code change should be introduced.

I'm not happy with the layout of "templating" but that seems to be an pre-existing issue?


marialungu (approved):
LGTM 🚀

PR #137, opened 2023-04-24
feat: use async errors

raed667:
To test locally, run `yarn playground:start` and then visit `http://localhost:1234/error`

image


packages/recommend-react/src/useRecommendations.ts
```diff
@@ -49,10 +51,12 @@ export function useRecommendations<TObject>({
...
-    }).then((response) => {
-      setResult(response);
-      setStatus('idle');
-    });
+    })
+      .then((response) => {
+        setResult(response);
+        setStatus('idle');
+      })
+      .catch(throwAsyncError);
```
Haroenv, line 59: pretty sure this is a breaking change, no?

raed667: Indeed it is. This draft-PR is more as something we can link to in this error management discussion, rather than something I want to open right away.

examples/demo/src/routes/HomePage.tsx
```diff
@@ -24,27 +25,30 @@ export const HomePage: React.FC = () => {
...
-          />
-        )}
-        maxRecommendations={5}
-        translations={{
-          title: 'Trending Facet (brand)',
-        }}
-      />
+      <ErrorBoundary>
+        <TrendingFacets<FacetHit>
+          // recommendClient={recommendClient}
```
raed667:
```suggestion
```

PR #140, opened 2023-06-12
fix: remove vdom renderer from react props

raed667:
As internally reported by @sarahdayan 

> There seems to be an issue with the types in @algolia/recommend-react for the view prop in the RelatedProducts component.
It requires a renderer (createElement and Fragment) which seems to be inherited from the underlying VDOM package (which makes sense) but in React the renderer shouldn't be required since it's provided.

The proposed fix is to omit `Renderer`[props](https://github.com/algolia/recommend/blob/next/packages/recommend-vdom/src/types/Renderer.ts#L26) from the react components.

### DIFF
--- a/packages/recommend-react/src/FrequentlyBoughtTogether.tsx
+++ b/packages/recommend-react/src/FrequentlyBoughtTogether.tsx
@@ -17,7 +17,10 @@ const UncontrolledFrequentlyBoughtTogether = createFrequentlyBoughtTogetherCompo
 export type FrequentlyBoughtTogetherProps<
   TObject
 > = GetFrequentlyBoughtTogetherProps<TObject> &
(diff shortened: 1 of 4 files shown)


francoischalifour (approved):

sarahdayan (approved):
This looks fine to me!

You could potentially create a `WithoutRenderer` helper type which should be helpful as you add new components:

```ts
type WithoutRenderer<TType> = Omit<TType, 'createElement' | 'Fragment'>
```

marialungu (approved):

raed667:
@sarahdayan excellent suggestion ! 👍 

I'll publish this fix for now, and do a follow-up PR

francoischalifour:
We could even create the type to apply to all components props:

```ts
type ExtractComponentProps<TType> = Omit<
  TType,
  'items' | 'status' | 'createElement' | 'Fragment'
>
```

PR #148, opened 2023-08-02
test: verify that umd bundles load successfully

dhayab:
**Summary**

This PR (with similar ones for instantsearch and autocomplete) ensures that our umd bundles load and define their global variable successfully.

**Motivation**

We recently had an issue with the UMD bundle of InstantSearch.js resulting in the browser being unable to load the library. This was caused by a dependency that was not correctly transpiled, leaking commonjs declarations that the browser could not understand.


raed667 (approved):
✨✨✨ 💖 👍  WUNDERBAR 👍  💖 ✨✨✨

PR #154, opened 2023-09-28
fix: Fix @algolia/recommend-react package description

raed667:


francoischalifour:
Oops. Also looks like we don't have any descriptions for the other packages.

raed667:
@francoischalifour indeed ! I didn't want to editorialize, but it would be nice to do a PR and patch release to NPM so we can have all the packages with proper descriptions

marialungu (approved):

PR #155, opened 2023-10-23
feat(demo): add comparison chart example

marialungu:
Add implementation example for a comparison chart pattern
Screenshot 2023-10-20 at 16 30 50


examples/demo/src/components/ComparisonChartItem/index.tsx
```diff
@@ -0,0 +1,118 @@
...
+        >
+          <div className="uic-ComparisonChart-itemImageContainer">
+            <img
+              className="uic-ComparisonChart-itemImage"
+              src={item.image_urls[0]}
+              alt={item.name}
+            />
+          </div>
+        </a>
+        <button
```
raed667: For readability of the example, can you create a Button component to be used here?

marialungu: ✅

raed667 (approved):
🥇 LGTM 🥇

seafoox:
@marialungu great to see you're moving forward with the idea
* Assuming that the comparaison table is designed to take place on the PDP, do you confirm we'll also be showing as the first item, the currently viewed item? So customer can best compare the current one with alternatives products?
* Are you considering another UI iteration? Looking at what websites (eg. Amazon) do I believe we could enhance the legibility of the provided info.

seafoox:
Here some recommendations from Baymard on how to best design comparaison tables https://baymard.com/premium/guideline-collections/vnxizq/2142

PR #158, opened 2023-10-31
feat: add personalization to recommend calls

raed667:


examples/demo/src/components/BundleItem/index.tsx
```diff
@@ -24,6 +24,7 @@ export const BundleItem: React.FC<BundleItemProps<ProductHit>> = ({
 
         onSelect(item);
         insights('clickedObjectIDs', {
+          userToken: 'user-token-1',
```
raed667, line 27: Not sure how necessary this is since we're setting `userToken` on insights init

package.json
```diff
@@ -78,5 +78,8 @@
     "stream-http": "^3.1.0",
     "typescript": "4.9.5",
     "url": "^0.11.0"
+  },
+  "dependencies": {
+    "@algolia/client-personalization": "^4.20.0"
```
raed667: I'll remove this since using the HTTP endpoint was simpler for the PoC

PR #163, opened 2023-12-13
chore: update trending-facets types

raed667:
## 🤔 What 

Update `trending-facets` model types inherited from `@algolia/recommend` `v4.22.0` 

https://github.com/algolia/algoliasearch-client-javascript/releases/tag/4.22.0

## 🤷 Why 

The trending-facets API only returns type string for facetValue.

All other models return items/hits which are of type `RecordWithObjectID`

However `trending-facets` model returns a list of `TrendingFacetHit` which *(by definition)* don't have an `objectID` property.

To simplify the code, we can remove the type argument `TObject` for `trending-facets`

## 🔍 How to upgrade

If you're **not** using TypeScript, there is nothing to change. You can ignore the following.

If you're using TypeScript, and the `trending-facets` model,  you should update your code like the following, by removing the type argument passed: 

```ts
// The TrendingFacets React component usage goes from this:

type FacetHit = Hit<{
  facetName: string;
  facetValue: string;
}>
<TrendingFacets<FacetHit>
  indexName="my-index"
  // ...

// To this:
<TrendingFacets
  indexName="my-index"
  // ...

// or using hooks
const results = useTrendingFacets<FacetHit>(/* params */)
// becomes like this in v1.12.0
const results = useTrendingFacets(/* params */)
```
</details>

If you're using `@algolia/recommend-js` you can remove the `FacetHit` type in a similar way: 

```ts 

trendingFacets<FacetHit>({
  container: '#trendingFacets',
  indexName,
  // ...

// needs to be changed like this after v1.12.0: 
trendingFacets({
  container: '#trendingFacets',
  indexName,
  // ...
```
 
If  you encounter an issue with the upgrade, open an issue referencing this release.


marialungu (approved):

packages/recommend-core/src/getBatchRecommendations.ts
```diff
@@ -55,10 +56,18 @@ export async function getBatchRecommendations<TObject>({
...
-      recommendations = mapByScoreToRecommendations<TrendingFacet<TObject>>({
+    let recommendations: Array<ProductRecord<ProductRecord<TObject>>> = [];
+    let trendingFacets: Array<TrendingFacet<TFacet>> = [];
+    if (model === 'trending-facets') {
+      trendingFacets = mapByScoreToRecommendations<TrendingFacet<TFacet>>({
+        maxRecommendations,
+        hits: splitResult
+          .map((res) => res.hits)
+          .flat()
+          .map((hit) => (hit as unknown) as TrendingFacet<TFacet>),
```
raed667: not super happy about this line

examples/demo/src/types/FacetHit.ts
```diff
@@ -1,14 +1,3 @@
...
-};
-
-type WithInsights<THit> = THit & {
-  __position: string;
-  __indexName: string;
-  __queryID: string;
-};
-
-export type FacetHit = WithInsights<Hit<FacetRecord>>;
+export type FacetHit = TrendingFacetHit<string>;
```
francoischalifour: I don't think `TrendingFacetHit` is generic anymore?

examples/templating/createElement/index.ts
```diff
@@ -90,7 +90,7 @@ trendingItems<ProductHit>({
   },
 });
 
-trendingFacets({
+trendingFacets<string>({
```
francoischalifour: Don't we want to remove all generics for this model APIs?

packages/recommend-vdom/src/types/RecommendComponentProps.ts
```diff
@@ -15,7 +15,7 @@ export type HeaderComponentProps<TObject> = Renderer & ComponentProps<TObject>;
 
 export type ComponentProps<TObject> = {
   classNames: RecommendClassNames;
-  recommendations: TObject[];
+  recommendations: TObject[] | Array<FacetEntry<TObject>>;
```
francoischalifour: We should prob pass `Array<FacetEntry<TObject>>` at call site for `TObject`. Otherwise we'll have wrong types for other models which would assume that they can receive `Array<FacetEntry<TObject>>` recommendations.

packages/recommend-core/src/utils/mapByScoreToRecommendations.ts
```diff
@@ -1,21 +1,21 @@
...
-}: MapToRecommendations<TObject>): Array<ProductRecord<TObject>> {
+}: MapToRecommendations<THit>): THit[] {
   // Since recommendations from multiple indices are returned, we
   // need to sort them descending based on their score.
-  return sortBy<ProductRecord<TObject>>((a, b) => {
-    const scoreA = a._score || 0;
-    const scoreB = b._score || 0;
+  return sortBy<THit>((a, b) => {
+    const scoreA = a['_score'] ? a['_score'] : 0;
+    const scoreB = b['_score'] ? b['_score'] : 0;
```
francoischalifour: Isn't `_score` always returned by the Recommend API?

If not, perhaps we can change the generic to something like `THit extends { _score?: number }`.

packages/recommend-core/src/getBatchRecommendations.ts
francoischalifour: Do we even need to run this function on facets?

Anyway we shouldn't need to loop over the items to cast `hits`.

packages/recommend-vdom/src/types/RecommendComponentProps.ts
raed667: Good catch !

examples/demo/src/types/FacetHit.ts
raed667: Indeed, i'm currently making changes to use `v4.22.0` should have the PR ready soon.

examples/templating/createElement/index.ts
raed667: Indeed, i just pushed a commit to achieve that.

packages/recommend-core/src/getBatchRecommendations.ts
raed667: Good question 👀 indeed this is not useful since the goal of this function is to handle when we pass multiple queries to the component, however in our case we can't pass multiple facetNames ! 


Really good catch !

examples/js-demo/app.tsx
```diff
@@ -98,6 +100,26 @@ autocomplete<ProductHit>({
...
+    return <Facet hit={item} />;
+  },
+  view: (...props) => {
+    return (
+      <div className="Facets-Wrapper">
+        {props[0].items.map((item) => {
+          return <Facet key={item.facetValue} hit={item} />;
+        })}
+      </div>
+    );
```
raed667: Not 100% sure if `view` is meant to be used like this ? But I wanted to add a `trendingFacets` JS example.

francoischalifour: That seems fine but you can use `props` instead of `...props`.

packages/recommend-react/src/useFrequentlyBoughtTogether.ts
```diff
@@ -103,7 +105,7 @@ export function useFrequentlyBoughtTogether<TObject>({
   ]);
 
   return {
-    ...result,
+    recommendations,
```
francoischalifour: Why did we change all these returns? The reasoning was that if the API adds information in the response, we don't need to release another version of the library to leverage them.

packages/recommend-core/src/types/TrendingFacet.ts
```diff
@@ -1,5 +1,2 @@
-export type TrendingFacet<TObject> = TObject & {
-  _score?: number;
-  facetName?: string;
-  facetValue?: string;
-};
+import { TrendingFacetHit } from '@algolia/recommend';
+export type TrendingFacet = TrendingFacetHit;
```
francoischalifour: Do we still need this intermediary type?

raed667: good catch!  this was on my list of things to remove once the build passes !

The thing is that `vdom` imports the type from `core` (its sole dependency) so i'll need to modify  `vdom` to depend on `@algolia/recommend` as well, which i'm not sure of the impact.

packages/recommend-react/src/useFrequentlyBoughtTogether.ts
raed667: With Batching the typing becomes kind of chaotic around here, i'm open to suggestions

packages/recommend-react/src/useFrequentlyBoughtTogether.ts
```diff
@@ -74,7 +74,7 @@ export function useFrequentlyBoughtTogether<TObject>({
           setStatus('loading');
         },
         onResult(response) {
-          setResult(response as GetRecommendationsResult<TObject>);
+          setResult((response as unknown) as GetRecommendationsResult<TObject>);
```
raed667: @francoischalifour reset the state here to reflect your feedback, but i'm not super happy with the type. Any ideas on how we can improve this?  cc @marialungu

marialungu (approved):

packages/recommend-react/src/useFrequentlyBoughtTogether.ts
raed667: Fixed in latest commits

packages/recommend-react/src/useLookingSimilar.ts
```diff
@@ -29,7 +29,9 @@ export function useLookingSimilar<TObject>({
   const queryParameters = useStableValue(userQueryParameters);
   const fallbackParameters = useStableValue(userFallbackParameters);
 
-  const { hasProvider, register } = useRecommendContext();
+  const { hasProvider, register } = useRecommendContext<
+    GetRecommendationsResult<TObject>
```
francoischalifour, line 33: Couldn't this generic be passed to the `register()` call?

This line sounds like a lie. The Recommend context isn't of this type, you want to register a new model with this type.

We can further improve in follow up PRs.

packages/recommend-vdom/src/FacetsView.tsx
```diff
@@ -20,7 +19,7 @@ export function createFacetsView({ createElement, Fragment }: Renderer) {
         <ol className={cx('auc-Recommend-list', props.classNames.list)}>
           {props.items.map((item) => (
             <li
-              key={item.facetValue}
+              key={`${item.facetValue}`}
```
francoischalifour: Any reasons for this explicit casting?

packages/recommend-vdom/src/TrendingFacets.tsx
```diff
@@ -13,9 +13,7 @@ export function createTrendingFacetsComponent({
   createElement,
   Fragment,
 }: Renderer) {
-  return function TrendingFacets<TObject>(
-    props: TrendingComponentProps<TObject>
-  ) {
+  return function TrendingFacets(props: TrendingComponentProps<{}>) {
```
francoischalifour: This would be the equivalent to before, as the other generic is optional.

```suggestion
  return function TrendingFacets(props: TrendingComponentProps) {
```

francoischalifour (approved):

packages/recommend-react/src/useLookingSimilar.ts
raed667: That was my initial attempt, as it indeed makes more sense!

However, getting the generic type to "propagate" from `register` to the rest of the Recommend context wasn't super obvious after.

packages/recommend-vdom/src/FacetsView.tsx
raed667: Oups i should have reverted that after the trending-facets fix!

PR #167, opened 2024-01-03
feat: Add Personalized Recommendations (experimental) 

raed667:
## Personalized recommendations 

Closed in favor of https://github.com/algolia/recommend/pull/174


examples/demo/src/routes/HomePage.tsx
```diff
@@ -24,9 +23,10 @@ export const HomePage: React.FC = () => {
   ] = useApplicationContext();
 
   return (
-    <Recommend recommendClient={recommendClient}>
+    <>
```
raed667, line 26: Removed the Recommend context here (but is kept in `ProductPage` to showcase the usage with and without.

examples/demo/src/config.ts
```diff
@@ -1,3 +1,3 @@
 export const appId = '93MWK2GLFE';
-export const apiKey = '63a2f2cf276ced37f901d8612ce5b40c';
+export const apiKey = '9f51610affadbae8e687ce009418c497';
```
raed667, line 2: Needed a key with the required ACLs

packages/recommend-core/src/personalization/getStrategy.ts
```diff
@@ -0,0 +1,34 @@
...
+  region: string;
+};
+
+type StrategyResponse = {
+  facetsScoring: Array<{
+    facetName: string;
+    score: number;
+  }>;
+  personalizationImpact: number;
+};
```
bengreenbank, line 13: Just making sure that you don't need the [`eventsScoring` param](https://www.algolia.com/doc/rest-api/personalization/#method-param-eventsscoring)?

packages/recommend-core/src/personalization/getStrategy.ts
```diff
@@ -0,0 +1,34 @@
...
+
+export const getStrategy = async ({ region, apiKey, appId }: GetStrategy) => {
+  const response = await fetch(
+    `https://recommendation.${region}.algolia.com/1/strategies/personalization`,
```
bengreenbank: You should be able to get the perso strategy from the [same domain](https://www.algolia.com/doc/rest-api/personalization/#domain) you use to get the affinities (`https://personalization.${region}.algolia.com`).

bengreenbank (approved):
Looks good perso-wise! 🚀 I only reviewed how you're getting the affinities and strategy and then how you're computing the optional filters.

packages/recommend-core/src/personalization/getStrategy.ts
raed667: Indeed, since it is not used to compute the filters, I opted to ignore `eventsScoring` to keep the type a bit simpler.

packages/recommend-core/src/personalization/getStrategy.ts
raed667: I didn't know that ! Good catch !

raed667:
@bengreenbank thanks for the review. 

I just added some session storage to cache the strategy (30min) and profile (10min) when possible.

Does that sound reasonable to you? (check [commit]( https://github.com/algolia/recommend/pull/167/commits/528b99874b5d46e9eb09a7e32d887660cd0275d0) )

bengreenbank:
> @bengreenbank thanks for the review.
> 
> I just added some session storage to cache the strategy (30min) and profile (10min) when possible.
> 
> Does that sound reasonable to you? (check [commit](https://github.com/algolia/recommend/pull/167/commits/528b99874b5d46e9eb09a7e32d887660cd0275d0) )

No worries!

It's hard to know the perfect caching interval, but:
- the strategy is refreshed whenever the customer updates it. So it will depend on how often the customer wants to change their strategy. I would say absolutely no lower than every 5 minutes, because that's how often the strategy is sent from GCP GCS to the search engine - but most customers are going to keep it the same for a reasonably long time (days, weeks etc).
- as for the user profile, it's a similar story. The profiles are regenerated whenever a new event arrives for that user, so there's no perfect interval for all customers.

Overall what you have looks good for now, and if you come across any issues then feel free to tweak them.

marialungu (approved):

Haroenv:
Later, when it's not experimental, how would you want the option to be called? would perso be automatic?

raed667:
> Later, when it's not experimental, how would you want the option to be called? would perso be automatic?

@Haroenv good question! eventually, the feature will be implemented on the API and we will use whatever the API contract dictates. (we will for sure remove the log region param as it will be handled internally by the API)

francoischalifour:
I'm failing to see a migration path from experimental to stable that makes sense with this way of exposing personalization. (Also, the codebase gets quite convoluted for a temporary feature.)

Since this addition already requires customers to update their code, it would make sense to me to introduce a sub-package specifically for experimental personalization, which supports top-level `userToken` and `region` params.

(Unfortunately we cannot get rid of `region` because it would necessitate a backend architecture change requiring an API gateway. We can't afford to do this for Perso v1, but that seems appealing for NeuralPerso.)

```jsx
import {
  Recommend,
  RelatedProducts,
} from '@algolia/recommend-react/experimental-personalization';

import { indexName, recommendClient } from './lib';

function App() {
  return (
    <Recommend
      recommendClient={recommendClient}
      userToken="10291287.12091287"
      region="eu"
    >
      <RelatedProducts indexName={indexName} />
    </Recommend>
  );
}
```

This way, there would be a clear transition from experimental to stable: replace the `@algolia/recommend-react/experimental-personalization` import with `@algolia/recommend-react`.

(Note that the `userToken` can also be leveraged for a future Insights integration.)

packages/recommend-core/src/types/Personalization.ts
```diff
@@ -0,0 +1,4 @@
+export type Personalization = {
```
francoischalifour: Those are `PersonalizationProps`, we can be more explicit with the name.

packages/recommend-core/src/types/Personalization.ts
```diff
@@ -0,0 +1,4 @@
...
+  userToken?: string;
+  region?: 'us' | 'eu';
```
francoischalifour: These optional fields make the component typings wrong. It either no personalization props, or all personalization props. So it would be something like `Props | Props & PersonalizationProps`.

packages/recommend-core/src/personalization/getAffinities.ts
```diff
@@ -0,0 +1,41 @@
...
+        'X-Algolia-API-Key': apiKey,
+      },
+    }
+  );
+
+  if (!response.ok) {
+    throw new Error(
+      `Failed to fetch personalisation affinities. Status: ${response.status}`
+    );
+  }
```
francoischalifour, line 37: Do we really want to break the recommendations behavior on the end-user screen when we're unable to fetch their user profile?

We better fallback to non-personalized recommendations. That's similar to how the Search Engine deals with non-existant user tokens, and that's similar to the future Recommend backend behavior.

(Same comment for the strategy fetching.)

packages/recommend-core/src/personalization/index.ts
```diff
@@ -0,0 +1,42 @@
...
+
+export const getPersonalizationFilters = async ({
+  userToken,
+  region,
+  apiKey,
+  appId,
+}: GetPersonalizationFilters) => {
+  if (!userToken || !region || !apiKey || !appId) {
+    return [];
+  }
```
francoischalifour: Missing `region`, `apiKey` and `appId` are development errors, so they should be caught upfront and for good DX, should throw. (Ideally with `invariant()` but we don't have this dependency on this package yet.)

Missing `userToken` can be a development error or production scenario, so we could trigger a warning on top of failing silently.

packages/recommend-core/src/personalization/index.ts
```diff
@@ -0,0 +1,42 @@
...
+      getAffinities({ userToken, apiKey, appId, region }),
+      getStrategy({ apiKey, appId, region }),
+    ]);
+
+    const optionalFilters = Object.entries(affinities.scores).flatMap(
+      ([facet, values]) =>
+        Object.entries(values).map(([value, score]) => {
+          const weight =
+            strategy.facetsScoring.find((value) => value.facetName === facet)
+              ?.score || 100;
```
francoischalifour: Can we replace the magic number `100` with a named constant? This number represents the fallback weight, and by storing it as a constant, it clarifies its purpose and signals that it's open to future adjustments. (That's another discussion, but 100 is probably not a great default TBH.)

packages/recommend-core/src/getBatchRecommendations.ts
```diff
@@ -28,23 +30,73 @@ export type GetBatchRecommendations<TObject> = {
...
 
-  const response = await recommendClient.getRecommendations<TObject>(queries);
+  let _queries = queries;
+
+  /**
+   * Big block of duplicated code, but it is fine since it is experimental and will be ported to the API eventually.
+   * This is a temporary solution to get recommended personalization.
+   */
+  if (region && userToken) {
+    recommendClient.addAlgoliaAgent('personalization');
```
francoischalifour: We're tracking the experimentation here. (Needs to be updated in the other files too.)

```suggestion
    recommendClient.addAlgoliaAgent('experimental-personalization');
```
Or:

```suggestion
    recommendClient.addAlgoliaAgent('frontend-personalization');
```

packages/recommend-js/src/experimental-personalization/frequentlyBoughtTogether.tsx
```diff
@@ -0,0 +1,18 @@
...
+
+import { Personalization } from '@algolia/recommend-core';
+
+import {
+  FrequentlyBoughtTogetherProps,
+  frequentlyBoughtTogether as render,
+} from '../frequentlyBoughtTogether';
+import { EnvironmentProps, HTMLTemplate } from '../types';
+
+type Props<TObject> = FrequentlyBoughtTogetherProps<TObject, HTMLTemplate> &
```
francoischalifour: We export the props in the original package, but not in the experimental package. We better export them too so that the migration doesn't cause any mismatches.

packages/recommend-js/src/experimental-personalization/frequentlyBoughtTogether.tsx
```diff
@@ -0,0 +1,18 @@
...
+  EnvironmentProps &
+  Personalization;
+
+export function frequentlyBoughtTogether<TObject>(props: Props<TObject>) {
+  return render<TObject>(props);
+}
```
francoischalifour: Naming suggestions so that the names are clearer (no brand new names to learn, just add a suffix) and exported with correct naming.

```suggestion
import { Personalization } from '@algolia/recommend-core';

import {
  FrequentlyBoughtTogetherProps as FrequentlyBoughtTogetherPropsPrimitive,
  frequentlyBoughtTogether as frequentlyBoughtTogetherPrimitive,
} from '../frequentlyBoughtTogether';
import { EnvironmentProps, HTMLTemplate } from '../types';

type FrequentlyBoughtTogetherProps<TObject> =
  FrequentlyBoughtTogetherPropsPrimitive<TObject, HTMLTemplate> &
    EnvironmentProps &
    Personalization;

export function frequentlyBoughtTogether<TObject>(
  props: FrequentlyBoughtTogetherProps<TObject>
) {
  return frequentlyBoughtTogetherPrimitive<TObject>(props);
}
```

(Related to my top-level comment, but the code related to `experimental-personalization` should prob also live here.)

francoischalifour (commented):
All the experimental personalization code still lives in Recommend here. The goal would be for it to live in the experimental package and forwarded through `queryParameters`. Any reasons why it's not?

packages/recommend-core/src/personalization/index.ts
raed667: Good catch ! This is a merge issue ! 👀  i'll have to re-apply a commit where i did that

packages/recommend-core/src/getBatchRecommendations.ts
raed667: i like `experimental-personalization` 👍

packages/recommend-core/src/personalization/getAffinities.ts
raed667: There is a `try/catch` around these calls in `getPersonalizationFilters`. 

This throw here, is to check the HTTP status since fetch won't throw on `4xx` errors.

TL;DR when a userToken results in an HTTP error we fallback silently to non-personalised recommend.

packages/recommend-core/src/types/Personalization.ts
raed667: Yes ! i knew i forgot something ! Thanks !

packages/recommend-js/src/experimental-personalization/frequentlyBoughtTogether.tsx
raed667: Thanks ! i addressed this in the last commit 👍

packages/recommend-core/src/personalization/getAffinities.ts
francoischalifour: Indeed, I missed that

raed667:
Closed in favor of https://github.com/algolia/recommend/pull/174

PR #174, opened 2024-01-16
feat: Add Personalized Recommendations (experimental)

raed667:
## Personalized recommendations 

We want to add **experimental** support for personalized recommendations, using Algolia [Personalization](https://www.algolia.com/products/search-and-discovery/personalization/) "Classic" 

This is an experimental UI library implementation that stitches together existing APIs. 

A reference guide will also be shared for users who want to implement this in the flavor of their choice.

## How to use 

To enable personalisation with recommend, you'll need to import from `/experimental-personalization` like the following:

#### JavaScript 

```js
import { trendingItems } from '@algolia/recommend-js/dist/esm/experimental-personalization';

trendingItems({
  region: 'eu', // "eu" | "us"
  userToken: 'user_token',
  indexName: "indexName"
  // ...
```

Or by using a standalone endpoint: 
```
<script src="https://cdn.jsdelivr.net/npm/@algolia/recommend-js/dist/umd/experimental-personalization/index.js"></script>
```

#### React  

```jsx
import { TrendingItems } from '@algolia/recommend-react/dist/esm/experimental-personalization';

<TrendingItems
  region="eu" // "eu" | "us"
  userToken="user_token"
  indexName="indexName"
  // ...
```

Or by using a standalone endpoint: 
```
<script src="https://cdn.jsdelivr.net/npm/@algolia/recommend-react/dist/umd/experimental-personalization/index.js"></script>
```

We avoid using the pre-existing `queryParameters` for `userToken` to avoid unexpected side-effects as well as possible semantic-versioning headaches if/when personalized recommendations is supported by the Recommend API and the `experimental` parameter is removed. 

### Parameters

The following parameters are all required for personalization:

- `region` :  Personalization data is stored and processed in a specific region. You can check the region for your account in the [Algolia dashboard](https://dashboard.algolia.com/account/infrastructure/analytics/).
- `userToken` : Associates a certain user token with the current query [learn more](https://www.algolia.com/doc/api-reference/api-parameters/userToken/)
- `suppressExperimentalWarning`: (optional) boolean parameter to disable the experimental console warning message.
 

# Notes 

- When using `recommend-react`: if the `Recommend` context is used for batching, the `userToken` passed to the component take precedence over the one passed to the context. (ie if you pass 2 different user-tokens between the context provider and the recommend component, the one from the component is used)

- When using the `<Recommend>` React context provider, you will need to import all the components from `/experimental-personalization`.

- Caching is proposed in a stacked PR https://github.com/algolia/recommend/pull/175


packages/recommend-core/src/personalization/__tests__/index.test.ts
```diff
@@ -0,0 +1,140 @@
...
+        getPersonalizationFilters({
+          userToken: 'token',
+          region: '',
+          apiKey: '',
+          appId: '',
+        })
+      ).rejects.toThrow(
+        `[Algolia Recommend] parameters 'region', 'apiKey' and 'appId' are required to enable personalization.`
+      );
+    });
```
francoischalifour: The test name doesn't match its content.

packages/recommend-core/src/personalization/getAffinities.ts
```diff
@@ -0,0 +1,40 @@
...
+        'X-Algolia-Application-Id': appId,
+        'X-Algolia-API-Key': apiKey,
+      },
+    }
+  );
+
+  if (!response.ok) {
+    throw new Error(
+      `Failed to fetch personalisation affinities. Status: ${response.status}`
+    );
```
francoischalifour:
```suggestion
    throw new Error(
      `Failed to fetch personalization affinities. Status: ${response.status}`
    );
```

packages/recommend-core/src/personalization/getPersonalizationFilters.ts
```diff
@@ -0,0 +1,54 @@
...
+
+export const getPersonalizationFilters = async ({
+  userToken,
+  region,
+  apiKey,
+  appId,
+}: GetPersonalizationFilters) => {
+  if (!region || !apiKey || !appId) {
+    throw new Error(
+      `[Algolia Recommend] parameters 'region', 'apiKey' and 'appId' are required to enable personalization.`
```
francoischalifour:
```suggestion
      `[Algolia Recommend] Parameters `region`, `apiKey` and `appId` are required to enable personalization.`
```

francoischalifour: Although `apiKey` and `appId` aren't explicit params provided by the developer, they're inferred from the client. So the message is misleading. We could have another branch for those.

packages/recommend-core/src/personalization/getPersonalizationFilters.ts
```diff
@@ -0,0 +1,54 @@
...
+    );
+  }
+  if (!userToken) {
+    // eslint-disable-next-line no-console
+    console.warn(
+      `[Algolia Recommend] parameter 'userToken' is required to enable personalization.`
```
francoischalifour:
```suggestion
      `[Algolia Recommend] Personalization couldn't be enabled because `userToken` is missing. Falling back to non-personalized recommendations.`
```

packages/recommend-core/src/personalization/getPersonalizationFilters.ts
```diff
@@ -0,0 +1,54 @@
...
+          const weight = facetsScoringMap.get(facet) ?? FALLBACK_WEIGHT;
+          const weightedScore = Math.floor(score * (weight / 100));
+          return `${facet}:${value}<score=${weightedScore}>`;
+        })
+    );
+
+    return optionalFilters;
+  } catch (error) {
+    return [];
+  }
```
francoischalifour: Can we still `console.error` the error? (And add a test that it shows the error.)

packages/recommend-core/src/personalization/getPersonalizationFilters.ts
```diff
@@ -0,0 +1,54 @@
...
+    );
+
+    const optionalFilters = Object.entries(affinities.scores).flatMap(
+      ([facet, values]) =>
+        Object.entries(values).map(([value, score]) => {
+          const weight = facetsScoringMap.get(facet) ?? FALLBACK_WEIGHT;
+          const weightedScore = Math.floor(score * (weight / 100));
+          return `${facet}:${value}<score=${weightedScore}>`;
+        })
+    );
```
francoischalifour: Is the engine able to receive `X` filters? Isn't there a limit where it would throw?

packages/recommend-core/src/personalization/getStrategy.ts
```diff
@@ -0,0 +1,34 @@
...
+        'Content-Type': 'application/json',
+        'X-Algolia-Application-Id': appId,
+        'X-Algolia-API-Key': apiKey,
+      },
+    }
+  );
+
+  if (!response.ok) {
+    throw new Error(
+      `Failed to fetch personalisation strategy. Status: ${response.status}`
```
francoischalifour:
```suggestion
      `Failed to fetch personalization strategy. Status: ${response.status}`
```

packages/recommend-core/src/personalization/isPersonalized.ts
```diff
@@ -0,0 +1,8 @@
+import { PersonalizationProps } from './types';
+
+export const isPersonalized = (object: any): object is PersonalizationProps => {
+  return (
+    (object.region === 'eu' || object.region === 'us') &&
+    typeof object.userToken === 'string'
+  );
```
francoischalifour: The user intent to enable personalization is to provide values for `region` and `userToken`, but not necessarily correct values. Otherwise this mixes the feature activation with feature validation. So we should rather check whether these values are defined.

packages/recommend-core/src/personalization/isPersonalized.ts
```diff
@@ -0,0 +1,8 @@
+import { PersonalizationProps } from './types';
+
+export const isPersonalized = (object: any): object is PersonalizationProps => {
```
francoischalifour: Naming suggestion: `isPersonalizationEnabled`.

packages/recommend-js/src/experimental-personalization/frequentlyBoughtTogether.tsx
francoischalifour, line 1: Is there any reasons we can't rely on the original model function?

```tsx
return frequentlyBoughtTogetherPrimitive({
  ...props,
  queryParameters: {
    ...props.queryParameters,
    optionalFilters: [
      // ...
    ],
  },
});
```

packages/recommend-core/src/personalization/index.ts
francoischalifour, line 1: We can name the folder `experimental-personalization`, like in the other packages.

francoischalifour (commented):
We gonna need some tests for all components to make sure that

- the correct search parameters are sent
- the correct warnings trigger
- the correct errors trigger

packages/recommend-core/src/personalization/__tests__/index.test.ts
raed667: good catch

packages/recommend-core/src/personalization/getPersonalizationFilters.ts
raed667: If there is a reasonable limit, i haven't seen it yet.
I tired some queries with ~25 filters and the engine responded fine.

Before adding a limit, I guess we can look at strategies and see how many attributes are used in maximum cases.

packages/recommend-core/src/personalization/isPersonalized.ts
raed667: IMO this function is a type guard, so it should do its very best to guarantee the parameter object is of type `PersonalizationProps` 

If we check just for existence `region` and `userToken` it could provide false type guarantees to code that depend on it.

packages/recommend-core/src/getTrendingItems.ts
```diff
@@ -49,6 +49,10 @@ export function getTrendingItems<TObject>({
 
   recommendClient.addAlgoliaAgent('recommend-core', version);
 
+  if (!indexName) {
+    return Promise.resolve({ recommendations: [] });
+  }
```
raed667, line 54: We don't have an objectID in the case of `trending-items` so this is the next best thing.

packages/recommend-core/src/personalization/isPersonalized.ts
francoischalifour: It's used differently in 2 places though:

https://github.com/algolia/recommend/blob/b717770c11668d08fab1453de4c99edfbc12d07a/packages/recommend-js/src/experimental-personalization/frequentlyBoughtTogether.tsx#L41-L43

And in another function:

https://github.com/algolia/recommend/blob/b717770c11668d08fab1453de4c99edfbc12d07a/packages/recommend-react/src/experimental-personalization/Recommend.tsx#L91-L93

If the developer makes a mistake in the region, the code path for personalization will not be run. It kinda fails silently. Same would happen if a new region is introduced; the check in the library prevents it from running. We shouldn't have too many runtime type checks for search parameters, we should let the API throw if needed.

We could perhaps do:

```tsx
// usage
const { userToken, region } = validatePersonalizationProps(props)
```

```tsx
// implementation
export function isPersonalizationEnabled(props) {
  return (
    typeof props.region !== 'undefined' && typeof props.userToken !== 'undefined'
  );
}

export function getPersonalizationProps(props) {
  if (isPersonalizationEnabled(props)) {
    return {
      region: props.region,
      userToken: props.userToken,
    };
  }

  return {};
}
```

raed667: I like `getPersonalizationProps` removes redundant checks !

marialungu (approved):

marialungu (approved):

packages/recommend-js/src/frequentlyBoughtTogether.tsx
```diff
@@ -26,7 +26,7 @@ const UncontrolledFrequentlyBoughtTogether = createFrequentlyBoughtTogetherCompo
   }
 );
 
-function useFrequentlyBoughtTogether<TObject>(
+export function useFrequentlyBoughtTogether<TObject>(
```
francoischalifour, line 29: Just bringing awareness that this will now be exported from the package, which isn't part of the contract. (Could be fine as long as we don't document it, so we can remove it after the experimental package is removed.)

packages/recommend-react/src/Recommend.tsx
```diff
@@ -30,7 +30,7 @@ function isRegistered<TObject>(
   return Boolean(widgets.find((w) => dequal(w.param, param)));
 }
 
-function getCacheKey<TObject>(queries: Array<BatchQuery<TObject>>) {
+export function getCacheKey<TObject>(queries: Array<BatchQuery<TObject>>) {
```
francoischalifour: All these exports mean that users will be able to use these functions, which increases our API surface. We want to avoid that. Could be easily done by moving these functions in their own file.

packages/recommend-core/src/getFrequentlyBoughtTogether.ts
```diff
@@ -29,6 +29,10 @@ export function getFrequentlyBoughtTogether<TObject>({
 
   recommendClient.addAlgoliaAgent('recommend-core', version);
 
+  if (queries.length === 0) {
+    return Promise.resolve({ recommendations: [] });
+  }
```
francoischalifour, line 34: I guess that's fine but there's no way back after we release it this way. What used to be an API error now fails silently.

packages/recommend-react/src/Recommend.tsx
raed667: Good catch. Moved these functions into a dedicated helper file. 👍

examples/demo/src/routes/HomePage.tsx
```diff
@@ -1,10 +1,9 @@
 import algoliarecommend from '@algolia/recommend';
 import {
-  Recommend,
   TrendingFacets,
   TrendingItems,
   RecommendedForYou,
-} from '@algolia/recommend-react';
+} from '@algolia/recommend-react/dist/esm/experimental-personalization';
```
Haroenv:
```suggestion
} from '@algolia/recommend-react/experimental-personalization';
```

this would be possible with `exports` in package.json (but not sure if it's worth it of course)

packages/recommend-react/src/experimental-personalization/FrequentlyBoughtTogether.tsx
```diff
@@ -0,0 +1,45 @@
+import { PersonalizationProps } from '@algolia/recommend-core';
```
Haroenv, line 1: what do you think about adding a readme to this folder that explains what to remove once it's done?

examples/demo/src/routes/HomePage.tsx
```diff
@@ -24,8 +20,9 @@ export const HomePage: React.FC = () => {
   ] = useApplicationContext();
 
   return (
-    <Recommend recommendClient={recommendClient}>
+    <>
       <TrendingFacets
+        recommendClient={recommendClient}
```
Haroenv, line 25: does this have a reason to move the client out of the provider? Can you not mix Recommend from regular RecommendReact with the experimental one? Is there a good error message if it's mixed?

Haroenv (commented):
interesting approach, I think it works overall

packages/recommend-react/src/experimental-personalization/FrequentlyBoughtTogether.tsx
raed667: Theoretically everything will be removed, as this will be part of the regular API calls, and the `userToken` will be part of `queryParameters`

examples/demo/src/routes/HomePage.tsx
raed667: I just wanted to show 2 ways to use Recommend with the Context Provider (PDP) and without (HomePage).

examples/demo/src/routes/HomePage.tsx
raed667: Good catch, i thought about adding some config for it, but for a beta feature where we will guide the initial clients during the setup, i don't think it is really needed.

examples/demo/src/routes/HomePage.tsx
raed667: And yes if you're using the Recommend Context (like the PDP), you will need to import all components from `/experimental-personalization`

examples/demo/src/routes/HomePage.tsx
Haroenv: alternatively if you put it directly in a `experimental-personalization` folder instead of inside dist/esm, that should work too, but again, same considerations apply.

examples/demo/src/routes/HomePage.tsx
Haroenv: So is there a clear error if you use the regular provider with the experimental components? I didn't see it in the files.

raed667: I just added a note in the PR description (that will go also in the release note). Again this is something we can guide the beta clients through their implementation.

PR #192, opened 2024-12-03
feat: Add `queryID` & `postion` as a preperation for Recommend Analytics [RECO-2391]

raed667:
## What 

In @algolia/recommend-js & @algolia/recommend-react we want to allow users to be able to send insights events. 

For this to be possible we want each recommend widget to add the following data to the hits when possible:

```ts
__position?: number;
__indexName: string;
__queryID?: string;
```

Picking up the work done on #124 & #123


examples/demo/src/components/Facet/index.tsx
```diff
@@ -1,29 +1,21 @@
...
       className="Hit"
       onClick={(event) => {
         event.preventDefault();
-
         onSelect(hit);
-        insights('clickedFilters', {
-          index: indexName,
-          eventName: 'Facet Clicked',
-          filters: [`${hit.facetName}:${hit.facetValue}`],
-        });
```
raed667, line 26: There is no analytics for model trending-facets

packages/recommend-core/src/utils/mapToRecommendations.ts
```diff
@@ -22,11 +23,15 @@ export function mapToRecommendations<TObject>({
...
+  queryIDs,
 }: MapToRecommendations<TObject>): Array<ProductRecord<TObject>> {
   const indexTracker: IndexTracker = {};
 
-  hits.forEach((hitsArray) => {
+  hits.forEach((hitsArray, idx) => {
     hitsArray.forEach((hit, index) => {
+      if (queryIDs[idx]) {
+        hit.__queryID = queryIDs[idx];
+      }
```
dhayab, line 34: Nitpick: could you reuse `addQueryID()` here?

packages/recommend-core/src/getBatchRecommendations.ts
```diff
@@ -71,13 +75,18 @@ export async function getBatchRecommendations<TObject>({
...
     } else {
       recommendations = mapToRecommendations<ProductRecord<TObject>>({
         maxRecommendations,
         hits: splitResult.map((res) => res.hits),
         nrOfObjs: keyPair.value,
-      });
+        queryIDs: splitResult.map((res) => res.queryID),
+      })
+        .map((hit) => addIndexName(hit, indexName))
+        .map((hit, idx) => addAbsolutePosition(hit, idx));
```
dhayab, line 89: Nitpick: I doubt this will have a significant performance impact based on the limited number of recommendations usually returned but you could remove one `map` iteration and transform your hits in sequence the first iteration.

dhayab (approved):
I only have a couple of nitpicky remarks, other than that it works for me 😃

packages/recommend-core/src/getBatchRecommendations.ts
raed667: I also thought about this 🤔 but with `n=30` I thought readability took precedence over optimization WDYT ?

dhayab: Makes sense, I agree with your point.

marialungu (approved):
LGTM

PR #196, opened 2025-11-05
feat: add fallbackParameters support to FBT [RECO-2443]

raed667:
Add `fallbackParameters` support to Frequently bought together model


marialungu (approved):
🚀

